PR #5, opened 2018-12-12
Add honeybadger

jcoyne:


config/honeybadger.yml
```diff
@@ -0,0 +1,2 @@
+---
+api_key: '82fc99e1'
```
mjgiarlo: Should this be in `shared_configs`?

mjgiarlo: Created a new issue for this.

PR #51, opened 2019-01-14
Update to dor-services-client 1.0.0

### LINKED ISSUE #50: Switch to dor-services-client 1.0
- [ ] Change the exception that is handled: https://github.com/sul-dlss/workflow-server-rails/pull/49

mjgiarlo:
Fixes #50
Refs sul-dlss/dor-services-client#38

In so doing, rescue the NotFoundResponse instead of UnexpectedResponse. The former is thrown when an object is not found in DOR, which we would expect for an object that has not yet been registered. This allows truly unexpected responses to raise an un-rescued exception, causing a stack trace (which we want in this case to make us aware of error conditions we have not yet anticipated).

Also, remove the `tzinfo-data` dependency. This is only included in the bundle on environments we do not deploy in (Windows and JRuby), and in the meantime, its presence in the bundle causes every bundle command to include a warning message. We did not add this to the Gemfile; it was automatically put there by a Rails generator, which has no knowledge of our deployment environments.

This message can trip up shell aliases and functions that use bundler, such as grepping the bundle to sleuth out behavior within dependencies. So, remove it.


bin/fetch-versions
```diff
@@ -13,15 +13,15 @@ require 'dor/services/client'
...
+                                         username: ARGV[2],
+                                         password: ARGV[3])
 
 druids.each do |druid|
   druid = "druid:#{druid}" unless druid.start_with? 'druid:'
-  version = Dor::Services::Client.object(druid).current_version
+  version = client.object(druid).current_version
   puts "#{druid},#{version}"
-rescue Dor::Services::Client::UnexpectedResponse
+rescue Dor::Services::Client::UnexpectedResponse, Dor::Services::Client::NotFoundResponse
```
mjgiarlo: @jcoyne 	Is this right? I'm not sure which exceptions we want to `rescue` here.

jcoyne: I think only NotFound, that's why I raised the idea of having a separate exception in the first place.

mjgiarlo: Pushed.

PR #62, opened 2019-01-17
Add endpoint for get_objects_for_workstep

### LINKED ISSUE #59: Implement get_objects_for_workstep
https://github.com/sul-dlss/dor-workflow-service/blob/845edfe4160a165d62b06530138f71e5c816a5c9/lib/dor/services/workflow_service.rb#L268-L294

Used here: https://github.com/sul-dlss/robot-master/blob/440740b7ed1d490f7621fbd7fe37ad69b2b51fa5/lib/robot-master/workflow.rb#L169

jcoyne:
Fixes #59

### DIFF
--- a/app/controllers/workflow_queues_controller.rb
+++ b/app/controllers/workflow_queues_controller.rb
@@ -26,7 +26,8 @@ class WorkflowQueuesController < ApplicationController
                     .select(:druid)
 
     scopes = [waiting_scope] + completed_step_scopes
-    @objects = WorkflowStep.find_by_sql(*intersect(scopes)).pluck(:druid)
+    # Get the druids that belong in all (intersection) of the scopes (ActiveRecord::Relations)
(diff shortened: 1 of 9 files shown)


app/controllers/workflow_queues_controller.rb
```diff
@@ -22,12 +18,63 @@ def all_queued
     @workflow_steps = @workflow_steps.where(hours_ago)
   end
 
+  # Used by robot-master:
```
mjgiarlo: There's a lot going on here inside the controller. I have three reactions:

1. Do you feel this is the best place for the SQL/ARel code to live? (Could some of this live in the model?)
2. I'm surprised by how much of this you had to roll your own of. (Not criticism, just surprise at having to do manual query parsing and scope logic.)
2. Wherever this lives, could you add more documentation about what all is going on here?

jcoyne: 1. I extracted it to IntersectQuery
2. Me too.
3. Added

PR #91, opened 2019-01-24
Parse XML body on WorkflowsController#update requests

### LINKED ISSUE #88: `WorkflowsController#update` not parsing XML
The ETD robots send workflow updates via HTTP PUT, sending along an XML payload which it expects the server to parse. For example, the robot issues an HTTP PUT to `https://sul-lyberservices-test.stanford.edu/workflow/dor/objects/druid:pk473sh2989/workflows/etdSubmitWF/submit-marc`, sending the following payload:

```xml
<?xml version="1.0"?>
<process name="submit-marc" status="error" errorMessage="can't modify frozen String" errorText="sul-robots1-test.stanford.edu"/>
```

With headers of `{:content_type=>"application/xml"}`.

Instead of parsing that XML and storing the relevant bits in the database, `WorkflowsController#update` sees no `'current-status'` param and wrongly sets the workflow step as `completed`.

mjgiarlo:
This changes the behavior of the `WorkflowsController#update` method to be closer to that of the Java workflow service.

Fixes #88


app/controllers/workflows_controller.rb
```diff
@@ -35,11 +35,14 @@ def show
...
-      find_step_for_process.update(status: params['current-status'])
-    else
-      find_step_for_process.update(status: 'completed')
-    end
+    parser = ProcessParser.new(process_from_request_body)
+    step = find_step_for_process
+
+    return render plain: process_mismatch_error(parser), status: :bad_request if parser.process != params[:process]
+
+    return render plain: status_mismatch_error(step), status: :precondition_failed if params['current-status'] && step.status != params['current-status']
```
jcoyne: This is probably the wrong status code. It doesn’t match the spec https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412

app/services/workflow_parser.rb
```diff
@@ -28,13 +28,19 @@ def create_workflow_steps
     end
   end
```
jcoyne: Why does this become public?

app/controllers/workflows_controller.rb
mjgiarlo: Good call.

app/services/workflow_parser.rb
mjgiarlo: From a prior tweak that I backed away from. Will revert. Good eye!

app/controllers/workflows_controller.rb
mjgiarlo: @jcoyne How about 409 Conflict? 400 Bad Request? other?

mjgiarlo:
@jcoyne 	Changes made and pushed.

app/controllers/workflows_controller.rb
```diff
@@ -35,11 +35,14 @@ def show
   end
 
   def update
-    if params['current-status'].present?
-      find_step_for_process.update(status: params['current-status'])
-    else
-      find_step_for_process.update(status: 'completed')
-    end
+    parser = ProcessParser.new(process_from_request_body)
```
mjgiarlo: I don't love the shape of this method. Happy to hear ideas for how to clean it up or refactor further.

app/services/process_parser.rb
```diff
@@ -3,27 +3,65 @@
...
+      next if value.nil?
+
+      hash[attribute] = value
+    end
+    hash
   end
 
-  def name
-    process.attr('name')
+  def process
```
mjgiarlo: Renamed this method in order to align the names of methods in this class with the names of fields in the database. This is what allows us to call `step.update(parser.to_h)` in the workflows controller.

app/services/process_parser.rb
```diff
@@ -3,27 +3,65 @@
 ##
 # Parsing Process creation request
 class ProcessParser
-  attr_reader :process
+  attr_reader :process_element
+
+  PROCESS_ATTRIBUTES = %i[
```
mjgiarlo: I believe this is the full list of attributes from the Java-based workflow service.

spec/factories/workflow_steps.rb
```diff
@@ -6,7 +6,7 @@
       "druid:abc123#{n}"
     end
     datastream { 'accessionWF' }
-    process { 'proc' }
+    process { 'start-accession' }
```
mjgiarlo: Line up the factory with the XML fixture data.

app/controllers/workflows_controller.rb
```diff
@@ -35,11 +35,14 @@ def show
...
+    step = find_step_for_process
+
+    return render plain: process_mismatch_error(parser), status: :bad_request if parser.process != params[:process]
+
+    return render plain: status_mismatch_error(step), status: :conflict if params['current-status'] && step.status != params['current-status']
+
+    step.update(parser.to_h)
```
mjgiarlo: The prior implementation allowed PUT requests containing no XML body, which had the effect of setting the status to `completed`. Can we verify that use case? It no longer exists after this PR; we now require an XML payload to set workflow step attributes.

app/controllers/workflows_controller.rb
mjgiarlo: Went with 409 due to conflict in current status.

PR #95, opened 2019-01-25
Add some potential future improvements to the README

jcoyne:


README.md
```diff
@@ -26,3 +26,17 @@ GET http://localhost:3000/dor/objects/:druid/workflows/:workflows
 PUT http://localhost:3000/dor/objects/:druid/workflows/:workflows
 GET http://localhost:3000/workflow_archive
 ```
+
+
+## TODO
+
+### Remove the STOMP dependency
+It would be great if we could remove the dependency on STOMP messaging.
+One way we could do this is if we only allowing update messages to come from dor-services-app.
```
mjgiarlo: `s/allowing/allow/`

PR #118, opened 2019-03-01
Add github version badge

tallenaz:


jcoyne:
I'm not sure I understand the point of this. Isn't the version just the deployed hash code?

tallenaz:
This is to fulfill a DevOpsDocs requirement to set a badge for the application version where that means a release number. For example, the preservation_catalog badge is at 1.9.0

jcoyne:
@tallenaz We need to rethink that requirement.  We should look at the the git version as it's the most clear and correct indicator for what is actually released/deployed. I suspect this requirement was written before we had honeybadger telling the #deploy channel what is released and when.  How should I propose a change to that procedure?  Is a PR on DevOpsDocs appropriate?

tallenaz:
@jcoyne I think a DevOpsDocs PR should be part of the process, maybe with the requirement of a 3rd party review if we think more than the two of us should be involved in the rethinking. For myself, it doesn't seem too controversial, but there may be other opinions, etc.

jcoyne:
Ref: https://github.com/sul-dlss/DevOpsDocs/pull/251

ndushay:
can this be closed now, and the branch deleted?

tallenaz:
Sure, thanks for the bump.

PR #133, opened 2019-03-20
Migrate without updating

jcoyne:


db/schema.rb
```diff
@@ -32,7 +32,7 @@
     t.string "lane_id", default: "default", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.boolean "active_version", default: false
+    t.boolean "active_version"
```
justinlittman: Why did you remove the default?

jcoyne: Oops, fixed that.

PR #142, opened 2019-03-21
Close the client when we're done

jcoyne:


justinlittman:
Also, worth checking the connection? https://github.com/stompgem/stomp/blob/dev/examples/client_putget.rb#L34

jcoyne:
@justinlittman I don't think so.  The connection can only be closed once receive is called: https://github.com/stompgem/stomp/blob/dev/lib/stomp/connection.rb#L512

PR #154, opened 2019-03-22
Extract a WorkflowTemplateService

jcoyne:


jcoyne:
@mjgiarlo I don't think so, this code is already has coverage via the controller test and no new behaviors were added/removed.

mjgiarlo:
@jcoyne 	OK, makes sense.

PR #162, opened 2019-04-02
Add a workflow list for all repositories

### LINKED ISSUE #156: Create a route that shows workflows for all repositories.
Currently any sdr repository workflows are not being indexed as they used to. This is because the old Java service ignored the repository parameter: https://github.com/sul-dlss/workflow-service/blob/26aeb8d8ee0e392f33948a29e17bfd0e8d61d472/src/edu/stanford/sulair/dlss/dor/resource/WorkflowResource.java#L187-L200

We should make a new route `/objects/{druid}/workflows` that shows workflows for all repositories.  Then we should make the old route redirect to this new one.  We must keep the old route working (a redirect is fine) because all the repository objects have dsLocation encoded in them and we don't want to update them all.

jcoyne:
Before deploying this we need to see if our workflow proxy can handle the redirect correctly.

Fixes #156

### DIFF
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -23,12 +23,17 @@ class WorkflowsController < ApplicationController
 
   # Display all steps for all versions of a given object
   def index
-    object_steps = WorkflowStep.where(repository: params[:repo], druid: params[:druid])
+    return redirect_to "/objects/#{params[:druid]}/workflows" if params[:repo]
+
+    object_steps = WorkflowStep.where(druid: params[:druid])
                                .order(:workflow, created_at: :asc)
                                .group_by(&:workflow)
 
     @workflows = object_steps.map do |wf_name, workflow_steps|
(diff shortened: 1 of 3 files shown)


jcoyne:
The proxy is not re-writing the Location header on the redirect correctly. It redirects to https://lyberservices-test.stanford.edu/objects/druid:hx908xy6904/workflows

jcoyne:
@tallenaz @justinlittman and I examined these options:
1. Figuring out how to get lyberservices-prod to rewrite location headers
2. Updating consumers to point directly at the workflow service
3. Making the app aware that it is being sent through a proxy.

As the proxy has been a historical confusing place, creates an extra failure mode and often has been troublesome for ops, we determined that option 2 would be the best choice.  Tony identified that we would need to update the iptables on the workflow service because currently it only accepts connections from the proxy.

We identified these consumers that need to be added to iptables and reconfigured:
1. dor-services-app
1. dor-indexing-app
1. argo
1 robots

tallenaz:
workflow-service-stage is set up to allow the stage consumers.

PR #169, opened 2019-04-13
 Add a route that automatically enqueues jobs to Resque 

jcoyne:
In order to use this we'll have to add a method to dor-workflow-client. Then use that client in lyber-core, and update robot master such that it doesn't enqueue for these workflows.

This is helpful because currently robot-master hammers the workflow service with expensive requests for "what stuff is ready to be enqueued?"  The current process also introduces additional delays.

- [x] need to mark the steps as queued?
- [x] figure out how to disable specific workflows in robot-master
  - add the workflows to skip here: https://github.com/sul-dlss/robot-master/blob/master/bin/robot-download-workflows#L43-L47
- [ ] figure out how to begin a workflow.
- [ ] ensure that skip-queue steps are calling the next method.  e.g. `sdr-ingest-received` https://github.com/sul-dlss/preservation_robots/blob/1d870126fa06abafa4f38a4a6ee6f961744b115d/robots/preservation_ingest/complete_ingest.rb#L44

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,7 @@ Rails.application.routes.draw do
         # Create should be a POST, but this is what the Java WFS app did.
         put ':workflow', to: 'workflows#create'
         put ':workflow/:process', to: 'steps#update'
+        post ':workflow/:process/next', to: 'steps#next'
       end
     end
   end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -4,6 +4,7 @@ dor_services:
   password: 'password'
 
 enable_stomp: true
+enable_queuing: true
 
 messaging:
   uri: 'failover:(stomp+ssl://localhost:61612,stomp://remotehost:61613)'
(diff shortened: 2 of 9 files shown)


config/routes.rb
```diff
@@ -9,7 +9,8 @@
       collection do
         # Create should be a POST, but this is what the Java WFS app did.
         put ':workflow', to: 'workflows#create'
-        put ':workflow/:process', to: 'workflows#update'
+        put ':workflow/:process', to: 'steps#update'
+        post ':workflow/:process/next', to: 'steps#next'
```
mjgiarlo: PR is looking good so far. I'm curious about mixing `post` and `put` here. Does `steps#next` 
function more like a create operation? The methods look similar to me.

config/settings.yml
```diff
@@ -4,8 +4,12 @@ dor_services:
   password: 'password'
 
 enable_stomp: true
+enable_queuing: true
```
mjgiarlo: Can you say more about what this change adds?

jcoyne:
Superseded by #189

PR #180, opened 2019-04-26
Various changes to get tests to work locally, including pinning Ruby …

justinlittman:
…version in Dockerfile, downgrading Rails, and setting db to be configured by env variables.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -2,24 +2,18 @@ default: &default
   adapter: postgresql
   encoding: unicode
   database: "<%= ENV.fetch('DATABASE_NAME', 'workflow-server') %>"
-  username: "<%= ENV['DATABASE_USERNAME'] %>"
-  password: "<%= ENV['DATABASE_PASSWORD'] %>"
-  host: "<%= ENV['DATABASE_HOSTNAME'] %>"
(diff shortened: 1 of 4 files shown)


config/database.yml
```diff
@@ -11,15 +11,9 @@ default: &default
 
 development:
   <<: *default
-  username: postgres
-  password: sekret
-  host: localhost
-  
+
```
jcoyne: These previously had the values set so they could connect to the database at https://github.com/sul-dlss/workflow-server-rails/blob/master/docker-compose.yml#L30 without having to define any environment variables.

justinlittman: Except then `docker-compose run -e "RAILS_ENV=test" app rake spec` results in:
```
PG::ConnectionBad:
  could not connect to server: Connection refused
  	Is the server running on host "localhost" (127.0.0.1) and accepting
  	TCP/IP connections on port 5432?
  could not connect to server: Address not available
  	Is the server running on host "localhost" (::1) and accepting
  	TCP/IP connections on port 5432?
```

I suspect this is due to Docker networking. Let me set the values as defaults instead.

jcoyne:
Looks like rubocop is giving you some grief.

PR #189, opened 2019-05-03
Adds redis queueing of next steps.

justinlittman:


app/controllers/steps_controller.rb
```diff
@@ -14,8 +14,14 @@ def update
     return render plain: status_mismatch_error(step), status: :conflict if params['current-status'] && step.status != params['current-status']
 
     step.update(parser.to_h)
+
+    # TODO: Do we only want to do this for certain updates, e.g., when setting status to completed?
+    # Enqueue next steps
+    next_steps = NextStepService.for(step: step)
+    next_steps.each { |next_step| Queue.enqueue(next_step) }
```
justinlittman: Are there some steps that we don't want to enqueue?

jcoyne: Yes, if the template has skipQueue as an attribute I don't think we want to enqueue it.  I think NextStepService is only returning things that can be queued though: https://github.com/sul-dlss/workflow-server-rails/blob/master/app/services/next_step_service.rb#L28

Gemfile
```diff
@@ -13,9 +13,11 @@ gem 'config', '~> 1.7'
 gem 'dor-services-client', '~> 1.2'
 gem 'honeybadger', '~> 4.1'
 gem 'jbuilder', '~> 2.5'
+gem 'lyber-core', '>= 4.0.3'
```
jcoyne: Do you think we can inline the one function rather than pull in lyber-core?  Lyber-core is another dependency I'd like to kill off.

app/controllers/steps_controller.rb
```diff
@@ -48,6 +54,7 @@ def process_from_request_body
   end
 
   def current_version
-    ObjectVersionService.current_version(params[:druid])
+    # Providing the version as a param is for local testing without needing to run DOR services.
```
jcoyne: +1

jcoyne: I'd like us to move to this pattern in production too.

docker-compose.yml
```diff
@@ -13,7 +13,10 @@ services:
       - DATABASE_HOSTNAME=db
       - DATABASE_PORT=5432
       - SECRET_KEY_BASE="${SECRET_KEY_BASE}"
-    image: 'suldlss/workflow-server:latest-dev'
+#      - RAILS_ENV=development
+      - SETTINGS__ENABLE_STOMP=false
+      - SETTINGS__REDIS__HOSTNAME=redis
+#    image: 'suldlss/workflow-server:latest-dev'
```
jcoyne: remove these commented out lines

PR #191, opened 2019-05-08
Add the Resque web console

jcoyne:
We can't deploy this because currently access to this machine is restricted to only a few inbound ips.


justinlittman:
See instead https://github.com/sul-dlss/robot-console

PR #193, opened 2019-05-09
ETDs should have the first step completed automatically

jcoyne:
Previously we were skipping it but it was marked as waiting.

Fixes https://github.com/sul-dlss/hydra_etd/issues/206


jcoyne:
Oh, no! good catch.

jcoyne:
I've restored config.ru.

PR #197, opened 2019-05-13
Added rubocop performance per deprecation warning.

justinlittman:


jcoyne:
Do we care about this gem? I don’t feel like we need it.

justinlittman:
I defer to your judgment. If you feel it isn't necessary, go ahead and close this PR.

jcoyne:
I'm not overly concerned about micro-optimizations. I'd prefer we just have fewer dependencies.

PR #198, opened 2019-05-13
Adds version to appropriate routes avoiding current version lookup.

justinlittman:


jcoyne:
This seems like a rather substantial change.  What would you think about making this a versioned api, e.g. `/v2/:dor/object/:druid/versions/:version/....` ?   Then perhaps we can make version required in v2?

PR #207, opened 2019-05-21
When updating workflow step, updates most recent version rather than …

justinlittman:
…querying DOR for version.


jcoyne:
@justinlittman does some Hydrus change need to be deployed before we can merge this?

justinlittman:
Depends on https://github.com/sul-dlss/hydrus/pull/319 being deployed.

PR #210, opened 2019-05-23
Reduce the number of queues to workflow x lane

jcoyne:
Prior to deploying, we need to update and deploy configuration for:
- [x] was_robots (https://github.com/sul-dlss/was_robot_suite/pull/124)
- [x] gis_robots (https://github.com/sul-dlss/gis-robot-suite/pull/198)
- [x] preservation_robots (https://github.com/sul-dlss/preservation_robots/pull/138)
- [x] common-assembly (https://github.com/sul-dlss/common-accessioning/pull/327)

### DIFF
--- a/app/services/queue_service.rb
+++ b/app/services/queue_service.rb
@@ -17,8 +17,7 @@ class QueueService
 
   # Enqueue the provided step
   def enqueue
-    # Perform the enqueue to Resque
-    Resque.enqueue_to(queue_name.to_sym, class_name, step.druid)
+    Resque.enqueue_to(queue_name, class_name, step.druid)
     Rails.logger.debug "Enqueued #{class_name} for #{step.druid} to #{queue_name}"
 
     # Update status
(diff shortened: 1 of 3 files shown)


app/services/queue_service.rb
```diff
@@ -18,7 +18,7 @@ def initialize(step)
   # Enqueue the provided step
   def enqueue
     # Perform the enqueue to Resque
```
ndushay: This comment seems ... a little confused and not helpful?

app/services/queue_service.rb
```diff
@@ -36,10 +36,10 @@ def step_name
   # Generate the queue name from step
```
ndushay: should "from step" be removed in this PR?

jcoyne: it is still being derived from the current step object.

justinlittman:
This scheme should be described in README (either here or commons-accessioning).

PR #213, opened 2019-05-24
Adds new show route for workflow step.

### LINKED ISSUE #212: Create workflow step show endpoint
This will be required by robots that need to lookup the lane of current step when initiating new workflows.

justinlittman:
closes #212

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -30,6 +30,12 @@ class StepsController < ApplicationController
     head :no_content
   end
 
+  def show
+    @step = find_step_for_process
+
+    render plain: '', status: :not_found if @step.nil?
+  end
+
   private
 
   def process_mismatch_error(parser)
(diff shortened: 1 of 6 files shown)


spec/requests/workflows/show_step_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+RSpec.describe 'Show a workflow step for an object', type: :request do
+  let(:step) do
+    FactoryBot.create(:workflow_step,
+                      status: 'error',
+                      error_msg: 'Bang!',
+                      lifecycle: 'submitted')
+  end
+
+  # let(:client) { instance_double(Dor::Services::Client::Object, current_version: '1') }
```
jcoyne: Cruft?

justinlittman: Yes. Fixed. Also added test for 404.

app/controllers/steps_controller.rb
```diff
@@ -30,6 +30,12 @@ def destroy_all
     head :no_content
   end
 
+  def show
+    @step = find_step_for_process
+
+    render plain: '', status: :not_found if @step.nil?
```
jcoyne: In rails you could do `head :not_found` if you have no content to return also.

Issue #231, opened 2019-08-14
run sweeper for anything queued longer than 12 hours


ndushay:
Per storytime 2019-08-13:

> For queued items that get stuck, there is a sweeper that runs every 24 hours which sets  items queued for longer than 24 hours back to waiting, but this may be too long for time sensitive items like ETDs and Hydrus

> Proposal: change sweeper to check for items queued longer than 12 hours, rather than 24.

https://github.com/sul-dlss/workflow-server-rails/blob/602c35ae037de7e49f366773e64855d88fe5db76/app/services/sweeper.rb#L22

Looks like it checks every 60 min:  https://github.com/sul-dlss/workflow-server-rails/blob/602c35ae037de7e49f366773e64855d88fe5db76/config/schedule.rb#L24 -- leaving this as is for now.

ndushay:
by the time I wrote this ticket, I had practically done the work ... so I just did the work, too.

ndushay:
Tests were a little trickier than I expected, but #237 is merged to master and Andrew notified, so closing this.

PR #236, opened 2019-08-16
README: update testing instructions

ndushay:
While working on #231, I was unable to run the tests locally per the previous README instructions, so I updated that section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,16 +32,25 @@ $ docker exec -it (container id) /bin/sh
 
 ## Testing:
 
-Note that we create the docker container without the test group, so to run tests locally, you will need
-- postgres running (`pg_ctl -D /usr/local/var/postgres start` works on my mac)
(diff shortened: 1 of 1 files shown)


ndushay:
Hrmm.  it didn't work for me, maybe because I followed the instructions: 

`docker-compose run -e "RAILS_ENV=test" app rake spec`  which didn't work as the test gems aren't in the docker container.

And when I had the docker container with postgres running, running the specs outside of docker didn't work for me.  Possibly due to a postgres permissions error, I dunno.

What would you like the README to say?    What was there is incorrect;  you aren't comfy with my changes ... so what do you think is the middle ground here?

jcoyne:
I'm saying that you don't have to start postgres locally, you can use the docker container:

```
docker-compose up db
bundle exec rspec
```

PR #271, opened 2019-10-31
Add callbacks for dor-services-app to use

jcoyne:
## Why was this change made?
This allows us to avoid polling for success

To be deployed with https://github.com/sul-dlss/workflow-server-rails/pull/271

### DIFF
--- a/config/workflows/dor/accessionWF.xml
+++ b/config/workflows/dor/accessionWF.xml
@@ -27,27 +27,34 @@
     <label>Shelve content in Digital Stacks</label>
     <prereq>remediate-object</prereq>
   </process>
-  <process batch-limit="100" error-limit="10" lifecycle="published" name="publish" sequence="8">
+  <process batch-limit="1" error-limit="5" name="shelve-complete" sequence="8" skip-queue="true">
+    <label>Signal that object has been shelved</label>
+  </process>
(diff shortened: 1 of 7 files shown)


config/workflows/dor/accessionWF.xml
```diff
@@ -27,27 +27,34 @@
...
+  <process batch-limit="1" error-limit="5" name="sdr-ingest-transfer" sequence="12">
     <label>Initiate Ingest into Preservation</label>
     <prereq>provenance-metadata</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" lifecycle="deposited" name="sdr-ingest-received" sequence="11" skip-queue="true">
+  <process batch-limit="1" error-limit="5" lifecycle="deposited" name="sdr-ingest-received" sequence="13" skip-queue="true">
     <label>Signal from SDR that object has been received</label>
   </process>
-  <process batch-limit="1" error-limit="5" name="reset-workspace" sequence="12">
+  <process batch-limit="1" error-limit="5" name="reset-workspace" sequence="14">
```
mjgiarlo: Looks like both `reset-workspace` and `end-accession` are `sequence="14"`. Intentional or nah?

jcoyne: nope. I'll fix.

PR #278, opened 2019-11-13
Stop using the repository field

jcoyne:
## Why was this change made?

`workflow` is sufficient as an identifier, we don't need an extra field.  The database is very big so this will reduce the size of the stored data and indexes.

n/a


ndushay:
is it possible we use "sdr" for preservation and "dor" for everything else, for historical reasons?  I don't see any particular reason to continue but wonder if there may be implications.

jcoyne:
@ndushay yes, historically that is what we did, but there's absolutely no reason that the workflow service needs to know about that.  As far as it is concerned, there is a workflow, which is identified by name.   Having more than that just adds complexity and bloats the database.  Earlier today we ran into a bug that we introduced because we set the wrong value for repository, this could have been avoided if we never needed to set it at all.

jcoyne:
@jmartin-sul I think https://github.com/sul-dlss/workflow-server-rails/pull/286 and https://github.com/sul-dlss/dor-workflow-client/pull/132 can help us sneak up on this change.

PR #279, opened 2019-11-13
add validation for workflow step status

peetucket:
## Why was this change made?

validate that the workflow step status exists and is an allowed value so that it is impossible to set a step to an invalid state or to leave the state out

QUESTION: are there other states that we need to add to the allowed list?

### DIFF
--- a/app/models/workflow_step.rb
+++ b/app/models/workflow_step.rb
@@ -7,6 +7,7 @@ class WorkflowStep < ApplicationRecord
   validates :process, presence: true
   validates :version, presence: true
   validates :repository, presence: true
+  validates :status, inclusion: { in: %w[waiting completed queued error skipped] }
 
   scope :lifecycle, -> { where.not(lifecycle: nil) }
   scope :incomplete, -> { where.not(status: %w[completed skipped]) }
--- a/spec/models/workflow_step_spec.rb
+++ b/spec/models/workflow_step_spec.rb
@@ -22,6 +22,18 @@ RSpec.describe WorkflowStep do
       expect(subject.valid?).to be false
     end
   end
+  context 'without valid status' do
+    it 'is not valid if the status is nil' do
+      expect(subject.valid?).to be true
+      subject.status = nil
+      expect(subject.valid?).to be false
+    end
+    it 'is not valid if the status is a bogus value' do
+      expect(subject.valid?).to be true
+      subject.status = 'bogus'
+      expect(subject.valid?).to be false
+    end
+  end
   describe '#as_milestone' do
     builder = {}
     before do
--- a/spec/controllers/steps_controller_spec.rb
+++ b/spec/controllers/steps_controller_spec.rb
@@ -20,14 +20,15 @@ RSpec.describe StepsController do
 
   describe 'PUT update' do
     context 'when updating a step' do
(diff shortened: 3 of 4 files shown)


app/models/workflow_step.rb
```diff
@@ -26,6 +30,10 @@ def as_milestone(xml)
                   version: version)
   end
 
+  def valid_status?
+    errors.add(:status, :invalid_status) unless VALID_STATUS.include? status&.downcase
```
jcoyne: why are we downcasing?

app/models/workflow_step.rb
```diff
@@ -7,6 +7,10 @@ class WorkflowStep < ApplicationRecord
   validates :process, presence: true
   validates :version, presence: true
   validates :repository, presence: true
+  validates :status, presence: true
+  validate :valid_status?
+
+  VALID_STATUS = %w[waiting completed queued error skipped].freeze
```
jcoyne: do we need `waiting`?  Last time I checked we don't use this.

app/models/workflow_step.rb
```diff
@@ -7,6 +7,10 @@ class WorkflowStep < ApplicationRecord
   validates :process, presence: true
   validates :version, presence: true
   validates :repository, presence: true
+  validates :status, presence: true
+  validate :valid_status?
```
jcoyne: Use inclusion instead: https://guides.rubyonrails.org/active_record_validations.html#inclusion

spec/models/workflow_step_spec.rb
```diff
@@ -22,6 +22,22 @@
...
+      expect(subject.valid?).to be false
+    end
+    it 'is not valid if the status is a bogus value' do
+      subject.status = 'bogus'
+      expect(subject.valid?).to be false
+    end
+  end
+  context 'status is case insensitive' do
+    it 'is valid if the status is valid but all caps' do
+      subject.status = 'COMPLETED'
```
jcoyne: I don't understand why we'd want this to be valid.

app/models/workflow_step.rb
peetucket: no particular reason, we just validate as requiring downcased values only

app/models/workflow_step.rb
peetucket: i think so -- isn't that the default state for new workflow steps?

app/models/workflow_step.rb
peetucket: ok

spec/models/workflow_step_spec.rb
peetucket: BECAUSE SOMETIMES PEOPLE LIKE TO SHOUT?

but seriously, i don't think its a requirement

mjgiarlo: :hear_no_evil: :laughing:

app/models/workflow_step.rb
jcoyne: Sorry, that's right. I was thinking of "hold".

jcoyne:
Looks good, but there are some tests that need fixing.

peetucket:
> Looks good, but there are some tests that need fixing.

yeah, some of the other tests were using invalid statues, causing them to start failing --- fixed it all

PR #280, opened 2019-11-14
Remove the repository column 

jcoyne:
Depends on #278 

## Why was this change made?
This will reclaim a bunch of index space and make inserts faster (fewer indexes to update)

n/a


ndushay:
tests not passing

PR #281, opened 2019-11-14
validate version in workflow step

peetucket:
## Why was this change made?

just continue down the path of validating step data so we can't create bad data

### DIFF
--- a/spec/controllers/steps_controller_spec.rb
+++ b/spec/controllers/steps_controller_spec.rb
@@ -25,7 +25,6 @@ RSpec.describe StepsController do
       it 'updates the step with repository (Deprecated)' do
         put :update, body: body, params: { repo: repository, druid: druid, workflow: workflow_id,
                                            process: 'descriptive-metadata', format: :xml }
-        puts response.body
         expect(response.body).to eq('{"next_steps":[]}')
         expect(SendUpdateMessage).to have_received(:publish).with(druid: druid)
         expect(WorkflowStep.find_by(druid: druid, process: 'descriptive-metadata').status).to eq('error')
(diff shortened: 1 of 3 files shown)


spec/controllers/steps_controller_spec.rb
```diff
@@ -25,7 +25,6 @@
       it 'updates the step with repository (Deprecated)' do
         put :update, body: body, params: { repo: repository, druid: druid, workflow: workflow_id,
                                            process: 'descriptive-metadata', format: :xml }
-        puts response.body
```
peetucket: oops, left over from last PR debugging ... remove this

spec/models/workflow_step_spec.rb
```diff
@@ -34,6 +34,22 @@
       expect(subject.valid?).to be false
     end
   end
+  context 'without valid version number' do
+    it 'is not valid if the version number is nil' do
+      expect(subject.valid?).to be true
```
jcoyne: why bother checking this?

spec/models/workflow_step_spec.rb
```diff
@@ -34,6 +34,22 @@
...
+      subject.version = nil
+      expect(subject.valid?).to be false
+    end
+    it 'is not valid if the version number is not an integer' do
+      expect(subject.valid?).to be true
```
jcoyne: and this too?

spec/models/workflow_step_spec.rb
peetucket: no particular reason - i suppose this is just double-confirming the existing rails test for their validation

spec/models/workflow_step_spec.rb
peetucket: ditto above

spec/models/workflow_step_spec.rb
jcoyne: but that wouldn't fall under "without a valid version number".

peetucket: i guess i meant "version number" to generically mean the "version" attribute and not specifically a description of the data type

PR #40, opened 2019-01-10
add a unique constraint

jcoyne:

PR #283, opened 2019-11-15
add model level validation for uniqueness of step in a given workflow and version

### LINKED ISSUE #282: Add rails model level validation to ensure that we do not have duplicate rows for the same workflow/step/version
This will allow us to prevent new bad data from being added where we have duplicate rows for the same druid/workflow/step/version combination even if we are not yet able to add a database constraint (due to existing bad data) - see #40 for the database constraint work

This validation can be added now even with bad data and will help debug where the bad data is coming from.

peetucket:
## Why was this change made?

fixes #282 

note: we should eventually follow this up with the work in #40, but this is a good first step that can be implemented right now

### DIFF
--- a/spec/requests/workflows/destroy_all_spec.rb
+++ b/spec/requests/workflows/destroy_all_spec.rb
@@ -11,11 +11,12 @@ RSpec.describe 'Get a single workflow for an object', type: :request do
     let!(:item2) do
       FactoryBot.create(:workflow_step,
                         repository: 'sdr',
+                        process: 'publish',
                         druid: druid)
     end
 
     let!(:item3) do
-      FactoryBot.create(:workflow_step)
+      FactoryBot.create(:workflow_step, process: 'shelve')
     end
 
     let(:druid) { item1.druid }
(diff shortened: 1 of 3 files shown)


spec/requests/workflows/destroy_all_spec.rb
```diff
@@ -11,11 +11,12 @@
     let!(:item2) do
       FactoryBot.create(:workflow_step,
                         repository: 'sdr',
+                        process: 'publish',
```
peetucket: necessary for tests to continue to work with our new validation to prevent dupe steps from being added

spec/requests/workflows/destroy_all_spec.rb
```diff
@@ -11,11 +11,12 @@
...
                         druid: druid)
     end
 
     let!(:item3) do
-      FactoryBot.create(:workflow_step)
+      FactoryBot.create(:workflow_step, process: 'shelve')
```
peetucket: necessary for tests to continue to work with our new validation to prevent dupe steps from being added

PR #285, opened 2019-11-15
when creating workflow steps, ensure they have valid workflow name and process

peetucket:
## Why was this change made?

Prevent bad data from being created during remediation or other manual interventions by confirming that we can only create a new workflow row for an existing defined workflow, and that the step name provided is actually defined in the named workflow.  It is not necessary to perform this validation on update (and may be undesirable to do so) because it is theoretically possible for an existing workflow step name or workflow itself to change/be removed and still need to be updated later.

Note this validation is mostly there to protect from the sorts of manual remediations we are performing, as the creation of steps should in theory be undertaken only indirectly (when you ask the service to create an entire new workflow at once, where it then creates each step in process).

Note that there is a penalty cost to performing this validation when creating steps (namely, looking for the existence of the XML file and then parsing a struct) but I think it should be fast.

### DIFF
--- a/spec/requests/lanes_spec.rb
+++ b/spec/requests/lanes_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe 'Lanes', type: :request do
                       status: 'completed',
                       active_version: true)
     FactoryBot.create(:workflow_step,
-                      process: 'accept',
+                      process: 'shelve-complete',
                       lane_id: 'fast',
                       status: 'waiting',
                       active_version: true)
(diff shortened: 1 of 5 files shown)


spec/requests/lanes_spec.rb
```diff
@@ -20,7 +20,7 @@
                       status: 'completed',
                       active_version: true)
     FactoryBot.create(:workflow_step,
-                      process: 'accept',
+                      process: 'shelve-complete',
```
peetucket: these changes to process names (here and all the rest below) are necessary because this new validation requires valid step names, even in tests

PR #291, opened 2019-11-19
Distinguish between variables and subjects in RSpec

mjgiarlo:


mjgiarlo:
@jcoyne this is ready for another look (spun it into a new branch)

PR #293, opened 2019-11-25
Add a callback for preservation ingest

jcoyne:
## Why was this change made?

We will soon move SdrIngestTransfer to dor-services-app.  As this can take a long time, it will be performed in the background. This new workflow step will be able to log that that process completed successfully.

As nothing depends on this step (progress is blocked by sdr-ingest-received), this should not prevent any items from progressing through the workflow.  If we add this step to the template in prior to updating it (from dor-services-app), we will have to do less remediation of items currently in error somewhere in the workflow.

n/a

### DIFF
--- a/config/workflows/dor/accessionWF.xml
+++ b/config/workflows/dor/accessionWF.xml
@@ -45,16 +45,19 @@
     <label>Initiate Ingest into Preservation</label>
     <prereq>provenance-metadata</prereq>
   </process>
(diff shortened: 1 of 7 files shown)


mjgiarlo:
@jcoyne Tests need updating, it looks like.

jcoyne:
I'll deploy this for Andrew to test once he's finished up with https://github.com/sul-dlss/dor_indexing_app/pull/320

PR #295, opened 2019-12-05
Quiet honeybadger

mjgiarlo:
## Why was this change made?

Stop creating alerts for the unused repo parameter. This has occurred tens of thousands of times over the past couple weeks, and it is not clear to me how this alert is helping at this point.

no

### DIFF
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -42,7 +42,6 @@ class WorkflowsController < ApplicationController
   end
 
   # Display all steps for all workflows for all versions of a given object
-  # rubocop:disable Metrics/MethodLength
   def show
     workflow_props = { name: params[:workflow], druid: params[:druid] }
     scope = WorkflowStep.where(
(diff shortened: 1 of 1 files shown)


jmartin-sul:
> I'm in favor, but I think @jcoyne should sign off before merge. We do want to make sure we fix all the callers.

what if we put something in the logs that we could grep for? we'd still get the info we want, and while it'd be slightly less convenient, it'd be less of a strain on HB.

mjgiarlo:
@jcoyne same question as other HB-related one: what would you think of going with @jmartin-sul's idea of logging these to disk rather than to honeybadger?

jcoyne:
@mjgiarlo I rather prefer HB for this as it actually alerts us when these situations are encountered.  We don't have a mechanism for monitoring logs and sending alerts.

jcoyne:
@mjgiarlo I'd also like to point out that the rate of this error has already dropped dramatically. The last occurance was 5 days ago.

mjgiarlo:
OK, np.

PR #302, opened 2019-12-13
validate druids to have the correct format and prefix before creating

### LINKED ISSUE #300: Add validation for workflow step to ensure the druid has a 'druid:' prefix
Model object is not valid if druid is missing the druid: prefix.  This should help prevent things like this from happening: https://github.com/sul-dlss/preservation_catalog/issues/1267

peetucket:
## Why was this change made?

fixes #300 

Currently we have a bug in preservation_catalog where we are creating workflows without a druid prefix, and these results in workflows that aren't being picked up by indexing and shown in argo.  We should validate at this level to ensure this can't happen.

### DIFF
--- a/app/models/workflow_step.rb
+++ b/app/models/workflow_step.rb
@@ -2,7 +2,7 @@
 
 # Models a process that occurred for a digital object. Basically a log entry.
 class WorkflowStep < ApplicationRecord
-  validates :druid, format: { with: /\Adruid:[a-zA-Z]{2}\d{3}[a-zA-Z]{2}\d{4}/, message: 'is invalid format or missing prefix' }
+  validate :druid_is_valid
   validates :workflow, presence: true
   validates :process, presence: true
   validates :version, numericality: { only_integer: true }
@@ -31,6 +31,13 @@ class WorkflowStep < ApplicationRecord
   end
 
   ##
+  # check if we have a valid druid with prefix
+  # @return [boolean]
+  def valid_druid?
+    DruidTools::Druid.valid?(druid, true) && druid.starts_with?('druid:')
+  end
+
+  ##
   # check if the named workflow has a current definition
   # @return [boolean]
   def valid_workflow?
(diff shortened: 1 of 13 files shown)


spec/controllers/workflows_controller_spec.rb
```diff
@@ -31,7 +31,7 @@
   end
 
   describe 'deprecated PUT create' do
-    let(:druid) { 'druid:abc123' }
+    let(:druid) { 'druid:ab123ab1234' }
```
peetucket: note: had to fix dummy values in the tests so it wouldn't trigger the new validation

app/models/workflow_step.rb
```diff
@@ -2,7 +2,7 @@
 
 # Models a process that occurred for a digital object. Basically a log entry.
 class WorkflowStep < ApplicationRecord
-  validates :druid, presence: true
+  validates :druid, format: { with: /\Adruid:[a-zA-Z]{2}\d{3}[a-zA-Z]{2}\d{4}/, message: 'is invalid format or missing prefix' }
```
jmartin-sul: the extra dependency might be overkill, but seemed worth mentioning druid-tools as an option to keep from more copy/paste of this regex.  https://github.com/sul-dlss/druid-tools/#check-whether-a-druid-is-valid

mjgiarlo: Open to using the druidtools validator here? https://github.com/sul-dlss/druid-tools/blob/90be3a6fbeffe0be2822dd4ba9555ec80e4b6e12/README.md#check-whether-a-druid-is-valid

spec/controllers/workflows_controller_spec.rb
mjgiarlo: :sunglasses:

app/models/workflow_step.rb
peetucket: yeah, that works...still need to check for the prefix separately though, alas:

```
DruidTools::Druid.valid?('bb123cd4567')
==> true
```

peetucket: yes, that's fine, though will still need to check for a druid prefix, which DruidTools does not do

jmartin-sul: ah, good catch, didn't realize that

jmartin-sul:
@peetucket, i think feel free to merge yourself, assuming the rebase is uneventful

peetucket:
How do you feel about the new DruidTools commit vs the old approach?

jmartin-sul:
> How do you feel about the new DruidTools commit vs the old approach?

slightly prefer the `DruidTools` version, but don't feel strongly about it if you greatly prefer the original approach.

peetucket:
alrighty, rebased with the DruidTools approach, will leave for Monday AM...maybe Justin will weigh in

PR #304, opened 2019-12-16
remove unnecessary dependency

jcoyne:
## Why was this change made?
This prevents upgrading to rails 6

n/a


ndushay:
kewl.

PR #316, opened 2020-01-07
dependency updates

ndushay:
NOTE: please deploy this after merge is done; give headsup to Andrew first.

## Why was this change made?

to catch up on dep updates that weren't captured over our holidays. Since we're going to actively work in the code base, this seems prudent.

n/a


ndushay:
FYI:  Andrew and I just spoke:  he's okayed deployment of this and PR #314 (already merged) to stage and prod.

PR #325, opened 2020-01-23
Notify honeybadger when a version is not provided

jcoyne:
## Why was this change made?

We don't want the workflow service to call back to dor-services-app to get the version. The consumer should be providing it.  This allows us to identify consumers that are using the old behavior

n/a

### DIFF
--- a/spec/requests/lifecycle_spec.rb
+++ b/spec/requests/lifecycle_spec.rb
@@ -39,9 +39,19 @@ RSpec.describe 'Lifecycle', type: :request do
         allow(Dor::Services::Client).to receive(:object).with(druid).and_return(client)
       end
 
-      it 'draws an empty set of milestones' do
-        get "/dor/objects/#{druid}/lifecycle?active-only=true"
-        expect(returned_milestone_versions).to eq []
+      context 'when version is not passed (deprecated)' do
+        it 'draws an empty set of milestones and notifies honeybadger' do
+          expect(Honeybadger).to receive(:notify)
+          get "/dor/objects/#{druid}/lifecycle?active-only=true"
+          expect(returned_milestone_versions).to eq []
+        end
+      end
+
+      context 'when version is passed' do
+        it 'draws an empty set of milestones' do
+          get "/dor/objects/#{druid}/lifecycle?active-only=true&version=2"
+          expect(returned_milestone_versions).to eq []
+        end
       end
     end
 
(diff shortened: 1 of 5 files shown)


spec/requests/version_close_spec.rb
```diff
@@ -4,25 +4,41 @@
...
-  context 'deprecated route' do
+  context 'when version is not passed (deprecated)' do
+    let(:object_client) { instance_double(Dor::Services::Client::Object, version: version_client) }
+    let(:version_client) { instance_double(Dor::Services::Client::ObjectVersion, current: '2') }
+
+    before do
+      allow(Dor::Services::Client).to receive(:object).with(druid).and_return(object_client)
+    end
+
+    context 'with repository route (deprecated)' do
```
ndushay: should there be one of these clauses below?  asking for a friend.

spec/requests/workflows/create_workflow_spec.rb
```diff
@@ -24,11 +24,23 @@
...
         expect do
-          put "/#{repository}/objects/#{druid}/workflows/#{workflow}", params: request_data, headers: headers
+          put "/#{repository}/objects/#{druid}/workflows/#{workflow}?version=1", params: request_data, headers: headers
+        end.to change(WorkflowStep, :count).by(16)
+
+        expect(SendUpdateMessage).to have_received(:publish).with(druid: druid)
+      end
+    end
+
+    context 'when the version exists in dor-services-app (deprecated)' do
```
ndushay: add version is not passed, please.

spec/requests/workflows/create_workflow_spec.rb
```diff
@@ -70,39 +86,53 @@
...
-            post "/objects/#{druid}/workflows/#{workflow}"
+            post "/objects/#{druid}/workflows/#{workflow}?version=1"
           end.not_to change(WorkflowStep, :count)
           expect(response.status).to eq 400
         end
       end
     end
 
-    context "when the version doesn't exist" do
+    context 'when the version exists in dor-services-app (deprecated)' do
```
ndushay: add version is not passed, pls.

spec/requests/lifecycle_spec.rb
```diff
@@ -39,9 +39,19 @@
         allow(Dor::Services::Client).to receive(:object).with(druid).and_return(client)
       end
 
-      it 'draws an empty set of milestones' do
-        get "/dor/objects/#{druid}/lifecycle?active-only=true"
-        expect(returned_milestone_versions).to eq []
+      context 'when version is not passed (deprecated)' do
+        it 'draws an empty set of milestones' do
```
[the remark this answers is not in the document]
jcoyne: I don't mind changing the description, but I'm quite against negative assertions as there are an infinite number of negative assertions that one could make and they would all be true.

spec/requests/version_close_spec.rb
jcoyne: I'm not following what you are asking for.

spec/requests/lifecycle_spec.rb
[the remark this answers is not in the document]
ndushay: changing the descriptions works for me.

spec/requests/workflows/create_workflow_spec.rb
jcoyne: isn't line 27 taking care of that?

spec/requests/workflows/create_workflow_spec.rb
jcoyne: It wouldn't check with dor-services-app if the version were passed so that is implied.

spec/requests/version_close_spec.rb
ndushay: at line 38, shouldn't there be an outer clause stating it's without the repository route?

spec/requests/workflows/create_workflow_spec.rb
ndushay: in specs before this one, there is "when the version is passed" clause and then right after "when the version is not passed" clause, which makes it clear what is diff.

spec/requests/workflows/create_workflow_spec.rb
ndushay: okay, maybe for all these tests of deprecated ways, the tests will be going away once we know we don't need to keep the deprecated methods around, right?  So maybe it's not so important that these descriptions are clear to the folks that don't have all the pieces clear in their heads.

PR #354, opened 2020-03-30
Add lifecylce endpoint to non repo route

aaron-collier:
## Why was this change made?

The lifecycle route provides the information necessary to use the workflow-service for the milestone display in argo, however, it is not currently available without the repo parameter which we are trying to deprecate. 

README

No, not currently used.


aaron-collier:
@jcoyne updated the test to remove the `dor` references. Thanks.

PR #358, opened 2020-04-02
Ignore repository argument

mjgiarlo:
Connects to sul-dlss/preservation_robots#207 
Connects to sul-dlss/dor-workflow-client#159

## Why was this change made?

It enables dor-workflow-client to stop sending workflow steps qualified with (unused) repository names.

As part of this work, expand testing of repo-ful and repo-less interactions

No.

It makes sure the workflow service can operate when passed repository arguments and when not passed repository arguments.

### DIFF
--- a/app/controllers/workflow_queues_controller.rb
+++ b/app/controllers/workflow_queues_controller.rb
@@ -50,7 +50,7 @@ class WorkflowQueuesController < ApplicationController
   end
 
   def workflows_for_step_and_scope(step, scope)
-    _repository, workflow, process = step.split(':')
+    workflow, process = step.split(':').last(2)
 
     WorkflowStep.active.public_send(scope).where(workflow: workflow, process: process)
   end
--- a/spec/requests/lanes_spec.rb
+++ b/spec/requests/lanes_spec.rb
@@ -26,9 +26,23 @@ RSpec.describe 'Lanes', type: :request do
                       active_version: true)
   end
 
-  it 'shows the lanes' do
-    get '/workflow_queue/lane_ids?step=dor:accessionWF:shelve'
+  let(:expected_xml) do
+    '<lanes><lane id="default"/><lane id="fast"/></lanes>'
+  end
 
-    expect(response.body).to be_equivalent_to '<lanes><lane id="default"/><lane id="fast"/></lanes>'
+  context 'with repository-qualified step names' do
+    it 'shows the lanes' do
+      get '/workflow_queue/lane_ids?step=dor:accessionWF:shelve'
+
+      expect(response.body).to be_equivalent_to expected_xml
+    end
+  end
+
+  context 'without repository-qualified step names' do
+    it 'shows the lanes' do
+      get '/workflow_queue/lane_ids?step=accessionWF:shelve'
+
+      expect(response.body).to be_equivalent_to expected_xml
+    end
   end
 end
--- a/spec/requests/queues/all_queued_spec.rb
+++ b/spec/requests/queues/all_queued_spec.rb
@@ -16,6 +16,15 @@ RSpec.describe 'All queued steps', type: :request do
                       status: 'queued')
   end
 
+  let(:expected_xml) do
+    <<~XML
+      <workflows>
+        <workflow name="accessionWF" process="shelve" druid="#{one.druid}" laneId="default"/>
+        <workflow name="accessionWF" process="shelve-complete" druid="#{two.druid}" laneId="fast"/>
+      </workflows>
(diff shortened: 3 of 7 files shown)


app/controllers/workflow_queues_controller.rb
```diff
@@ -50,7 +50,7 @@ def find_completed_objects
   end
 
   def workflows_for_step_and_scope(step, scope)
-    _repository, workflow, process = step.split(':')
+    workflow, process = step.split(':').last(2)
```
mjgiarlo, line 53: This gives us more flexibility going forward.

If passed a step with a repository (*e.g.*, `dor:accessionWF:do-a-thing`), the repository is ignored and the workflow and process are set accordingly. In this way, the change is backwards-compatible with the former implementation. If, on the other hand, passed a step without a repository (*e.g.*, `accessionWF:do-a-thing`), this method will do the right thing: assign the workflow and step appropriately. This is an improvement over the past implementation and allows dor-workflow-client to begin phasing out the repository argument.

spec/requests/queues/all_queued_spec.rb
```diff
@@ -31,14 +31,29 @@
...
+  context 'with repository query arg' do
+    it 'shows all the queued items' do
+      get '/workflow_queue/all_queued?repository=dor&limit=3&hours-ago=24'
+
+      expect(response.body).to be_equivalent_to <<~XML
+        <workflows>
+          <workflow name="accessionWF" process="shelve" druid="#{one.druid}" laneId="default"/>
+          <workflow name="accessionWF" process="shelve-complete" druid="#{two.druid}" laneId="fast"/>
+        </workflows>
+      XML
```
ndushay, line 43: does it make sense for this to be in a let(:expected_xml) clause, partly to emphasis that the response is the same in both cases?

app/controllers/workflow_queues_controller.rb
ndushay: is it worth it / stupid to notify honeybadger if the workflow doesn't end in 'WF' ? or some such?   Just in case?

spec/requests/queues/objects_for_step_spec.rb
```diff
@@ -91,17 +91,34 @@
...
+          get '/workflow_queue?waiting=dor%3AaccessionWF%3Areset-workspace&' \
+              'completed=dor%3AaccessionWF%3Asdr-ingest-received&' \
+              'completed=dor%3AaccessionWF%3Aprovenance-metadata&limit=100&lane-id=default'
+
+          expect(response.body).to be_equivalent_to <<~XML
+            <objects count="2">
+              <object id="#{prereqs_and_waiting.druid}"/>
+              <object id="#{second_prereqs_and_waiting.druid}"/>
+            </objects>
+          XML
```
ndushay, line 105: ditto on expect_xml as memoized var?  and other places below.

spec/requests/queues/all_queued_spec.rb
mjgiarlo: dig it, will do

spec/requests/queues/objects_for_step_spec.rb
mjgiarlo: For sure.

app/controllers/workflow_queues_controller.rb
mjgiarlo: @ndushay I don't know that we've ever expressed a requirement that workflows must end in `'WF'`, so I'm meh on this change. If passed a bogus value for workflow or process, we'll get a HB alert eventually anyway :laughing:

mjgiarlo:
@ndushay Can you take another look? Thanks!

spec/requests/lanes_spec.rb
```diff
@@ -26,9 +26,19 @@
...
-    expect(response.body).to be_equivalent_to '<lanes><lane id="default"/><lane id="fast"/></lanes>'
+      expect(response.body).to be_equivalent_to '<lanes><lane id="default"/><lane id="fast"/></lanes>'
+    end
+  end
+
+  context 'without repository-qualified step names' do
+    it 'shows the lanes' do
+      get '/workflow_queue/lane_ids?step=accessionWF:shelve'
+
+      expect(response.body).to be_equivalent_to '<lanes><lane id="default"/><lane id="fast"/></lanes>'
```
ndushay, line 41: missed this one (I know it's short, but it helps conceptually that they are THE SAME.

mjgiarlo: got it!

PR #360, opened 2020-04-03
Fixed lane-id param when creating workflow.

### LINKED ISSUE #359: Lane not being set when workflows created
`lane-id` parameter is ignored.

See https://github.com/sul-dlss/workflow-server-rails/blob/master/app/controllers/workflows_controller.rb#L71

justinlittman:
closes #359

## Why was this change made?
So that the lane-id param matches the param passed by workflow client.

No.

If so, please confirm:
- change was tested on stage    and/or
- test added to sul-dlss/infrastructure-integration-test
Tested on stage.

### DIFF
--- a/app/services/process_parser.rb
+++ b/app/services/process_parser.rb
@@ -3,7 +3,7 @@
 ##
 # Parsing Process creation request
 class ProcessParser
-  attr_reader :process_element
+  attr_reader :process_element, :use_default_lane_id
 
   PROCESS_ATTRIBUTES = %i[
     process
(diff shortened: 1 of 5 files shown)


justinlittman:
image

app/services/process_parser.rb
```diff
@@ -24,8 +24,10 @@ class ProcessParser
 
   ##
   # @param [Nokogiri::XML::Element] process_element
-  def initialize(process_element)
+  # @param [Boolean] provide default as lane_id if no lane_id
+  def initialize(process_element, use_default_lane_id = true)
```
jcoyne, line 28: Would you mind making this a kwarg?  I find this makes the code much easier to read.

app/services/process_parser.rb
```diff
@@ -24,8 +24,10 @@ class ProcessParser
 
   ##
   # @param [Nokogiri::XML::Element] process_element
-  def initialize(process_element)
+  # @param [Boolean] provide default as lane_id if no lane_id
```
jcoyne, line 27: must add the parameter name (e.g. `use_default_lane_id`) here after `[Boolean]`

app/services/process_parser.rb
justinlittman: I do not mind. Fixed.

app/services/process_parser.rb
justinlittman: Fixed.

PR #362, opened 2020-04-06
update to dor-services-client v5

ndushay:
## Why was this change made?

So we can more easily upgrade all our apps at once to use the exact same version of cocina-models (which is pinned in dor-services-client)

na

na

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem 'rails', '~> 6.0.2'
 gem 'activerecord-import' # we can remove this after we've migrated the data
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'config', '~> 2.0'
-gem 'dor-services-client', '~> 3.9'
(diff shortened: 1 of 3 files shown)


ndushay:
will run integration tests on stage - deployed to stage already

ndushay:
```app/services/object_version_service.rb
19:    client.object(druid).version.current
26:  def client
27:    @client ||= Dor::Services::Client.configure(url: Settings.dor_services.url,

app/services/send_update_message.rb
30:    with_client do |client|
31:      client.publish(TOPIC_NAME, message.to_xml, headers)
43:  def with_client
44:    client = Stomp::Client.new(hosts: parse_failover_url(Settings.messaging.uri),
51:    yield client
52:    client.close
```

again - it's only using dsc for version ...

ndushay:
ran integration tests - all green.

PR #374, opened 2020-04-29
Added custom whenever job_type to remove deprecated warnings

jermnelson:
## Why was this change made?
To prevent Cron job created with whenever to send out warning emails and messages over depreciated calls with Ruby 2.7.1. 

n/a

If so, please confirm:
- change was tested on stage    and/or
- test added to sul-dlss/infrastructure-integration-test

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -20,7 +20,8 @@
 # end
 
 # Learn more: http://github.com/javan/whenever
+job_type :no_warnings_runner, "cd :path && RUBYOPT='-W0' bundle exec bin/rails runner -e :environment ':task' :output"
 
 every 60.minutes do
-  runner 'Sweeper.sweep'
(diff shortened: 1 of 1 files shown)


config/schedule.rb
```diff
@@ -20,7 +20,8 @@
 # end
 
 # Learn more: http://github.com/javan/whenever
+job_type :sweeper, "cd :path && RUBYOPT='-W0' bundle exec bin/rails runner -e :environment ':task' :output"
```
jermnelson, line 23: Decided to go with `RUBYOPT='-W0'` instead of `RUBYOPT='-W:no-deprecated'` in case this repo is deployed to production that may have an older Ruby version.

jcoyne, line 23: Can we change name to `no_warnings_runner`?

PR #377, opened 2020-05-06
Remove the completed steps

jcoyne:
## Why was this change made?

We're consolidating the shelve-complete and shelve steps as well as the publish and publish-complete steps.

Ref https://github.com/sul-dlss/common-accessioning/pull/656
Ref sul-dlss/argo#1789

no

Yes, was tested on stage.

### DIFF
--- a/config/workflows/accessionWF.xml
+++ b/config/workflows/accessionWF.xml
@@ -27,37 +27,31 @@
     <label>Shelve content in Digital Stacks</label>
     <prereq>remediate-object</prereq>
   </process>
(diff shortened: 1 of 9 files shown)


jcoyne:
@justinlittman I don't think so. What are you thinking?

justinlittman:
Will we have `*-complete` workflow steps that are eternally in `waiting`? Do we need to be concerned about workflows that are in the middle of/stuck at of a shelve or publish operation when this is deployed?

jcoyne:
Currently there are 3 items in shelve-complete with errors, I'll just move that status/error to the shelve step.

PR #381, opened 2020-05-13
[HOLD] Remove preservation-ingest-initiated step from accessionWF

jcoyne:
## Why was this change made?
Part of https://github.com/sul-dlss/argo/issues/1789

no

change was tested on stage

### DIFF
--- a/config/workflows/accessionWF.xml
+++ b/config/workflows/accessionWF.xml
@@ -39,19 +39,16 @@
     <label>Initiate Ingest into Preservation</label>
     <prereq>provenance-metadata</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" name="preservation-ingest-initiated" sequence="11" skip-queue="true">
-    <label>Signal that preservation ingest has been initiated</label>
-  </process>
-  <process batch-limit="1" error-limit="5" lifecycle="deposited" name="sdr-ingest-received" sequence="12" skip-queue="true">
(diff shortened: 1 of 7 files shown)


config/workflows/accessionWF.xml
```diff
@@ -39,19 +39,16 @@
     <label>Initiate Ingest into Preservation</label>
     <prereq>provenance-metadata</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" name="preservation-ingest-initiated" sequence="11" skip-queue="true">
-    <label>Signal that preservation ingest has been initiated</label>
-  </process>
-  <process batch-limit="1" error-limit="5" lifecycle="deposited" name="sdr-ingest-received" sequence="12" skip-queue="true">
+  <process batch-limit="1" error-limit="5" lifecycle="deposited" name="sdr-ingest-received" sequence="11" skip-queue="true">
```
ndushay, line 42: i strongly prefer "preservation" as it is less confusing than "sdr" which means a lot of different things.

ndushay: also - is this still "skip-queue?"

jcoyne: Changing the name of the existing step is out of scope.  no, it's not skip queue, we're just removing the one step and renumbering the remaining steps.

ndushay: this step you are keeping has "skip-queue" and i'm questioning if you still want it to say that.  I suspect you don't.

jcoyne: Yes, if that’s what it was, we want to keep it. We’re only changing the order attribute

PR #384, opened 2020-05-20
Added back published milestone.

justinlittman:
refs https://github.com/sul-dlss/argo/issues/2060

## Why was this change made?
So that the published lifecycle is provided in Argo display.

Stage

No


justinlittman:
image

PR #407, opened 2020-07-23
improve descriptions of etdSubmitWF steps.

ndushay:
## Why was this change made?

To get correct descriptions associated with etdSubmitWF

comments only

comments only

### DIFF
--- a/config/workflows/etdSubmitWF.xml
+++ b/config/workflows/etdSubmitWF.xml
@@ -13,19 +13,19 @@
     <label>Final Registrar approval (signal from PeopleSoft)</label>
   </process>
   <process name="submit-marc" sequence="5" skip-queue="true">
-    <label>Transfer stub MARC record to Symphony (drop off directory)</label>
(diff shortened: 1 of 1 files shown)


config/workflows/etdSubmitWF.xml
```diff
@@ -7,32 +7,32 @@
...
+    <label>Update descMetadata from Symphony, generate content metadata, rights metadata, create a tag.</label>
     <prereq>catalog-status</prereq>
   </process>
   <process name="start-accession" sequence="9" skip-queue="true">
     <label>Transfer control to Common Accessioning</label>
     <prereq>other-metadata</prereq>
   </process>
   <process name="binder-transfer" sequence="10" skip-queue="true" status="skipped">
-    <label>Transfer PDF to binder drop off directory for bound physical copy</label>
+    <label>No longer used.  (Transfer PDF to binder drop off directory for bound physical copy)</label>
```
mjgiarlo, line 36: I dare ask: should we drop this step then?

ndushay: no.  It's better to live junk around because of old data -- an object with an uncompleted WF causes problems later.

PR #410, opened 2020-08-05
use new completed_at date for milestones

peetucket:
## Why was this change made?

Fixes sul-dlss/argo#2192 ... basically because showing `updated_at` dates in argo for milestones is confusing, since those dates are bumped the the lifecycle is no longer active, hiding the dates when the steps were created.

I'm not sure if there are reasons we need to show `updated_at`, but I believe this data is only used for display purposes in Argo.

If we use a new `completed_at` date, this will only be triggered when a workflow step is completed.

New unit tests.

None.

### DIFF
--- a/app/models/workflow_step.rb
+++ b/app/models/workflow_step.rb
@@ -11,23 +11,49 @@ class WorkflowStep < ApplicationRecord
   validate  :workflow_exists, on: :create
   validate  :process_exists_for_workflow, on: :create
 
+  before_save :set_completed_at, if: :completed?
+
   scope :lifecycle, -> { where.not(lifecycle: nil) }
-  scope :incomplete, -> { where.not(status: %w[completed skipped]) }
-  scope :complete, -> { where(status: %w[completed skipped]) }
+  scope :incomplete, -> { where.not(status: COMPLETED_STATES) }
+  scope :complete, -> { where(status: COMPLETED_STATES) }
   scope :waiting, -> { where(status: 'waiting') }
   scope :queued, -> { where(status: 'queued') }
   scope :started, -> { where(status: 'started') }
   scope :active, -> { where(active_version: true) }
 
   scope :for_version, ->(version) { where(version: version) }
+
+  COMPLETED_STATES = %w[completed skipped].freeze # a list of states that are considered completed
   ##
   # Serialize a WorkflowStep as a milestone
   # @param [Nokogiri::XML::Builder] xml
   # @return [Nokogiri::XML::Builder::NodeBuilder]
   def as_milestone(xml)
     xml.milestone(lifecycle,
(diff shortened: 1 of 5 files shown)


app/models/workflow_step.rb
```diff
@@ -26,7 +26,7 @@ class WorkflowStep < ApplicationRecord
   # @return [Nokogiri::XML::Builder::NodeBuilder]
   def as_milestone(xml)
     xml.milestone(lifecycle,
-                  date: updated_at.to_time.iso8601,
+                  date: created_at.to_time.iso8601,
```
jcoyne, line 29: I believe the created_at timestamps get set when the workflow is initialized for an object.  This not the same as when the step is completed, which is what the updated_at gets.

jcoyne: How about we create a new column for "completed_at" that is only set when a process is completed.  The default value can be either updated_at or created_at.

peetucket: That would make sense.  And then the milestones API could return this "completed_at" date. This gets us away from the updated_at column, which can be inadvertently touched by rails when any other column is touched.

andrewjbtw: Sounds good to me. There are some steps where the completed_at will be noticeably later than the created_at when the object has a lot of data.

peetucket:
And this date change should just be for the life cycles view I believe. The workflows pop up window should remain as is I believe.

________________________________

peetucket:
I made a commit to do the `completed_at` idea. It includes a database migration to add a `completed_at` column and to set that column to the `updated_at` date by default for any existing completed steps.  A callback is used to set the `completed_at` date for new rows only when the step is completed.  Now for displaying milestone dates, the question is what to do want to return if the step is not yet completed?  In those cases, `completed_at` will be nil, so we can return the `created_at` date (what this PR currently does) or return a string (like 'n/a')?

app/models/workflow_step.rb
```diff
@@ -11,25 +11,51 @@ class WorkflowStep < ApplicationRecord
   validate  :workflow_exists, on: :create
   validate  :process_exists_for_workflow, on: :create
 
+  before_save :set_completed_at, if: :completed?
+
   scope :lifecycle, -> { where.not(lifecycle: nil) }
-  scope :incomplete, -> { where.not(status: %w[completed skipped]) }
-  scope :complete, -> { where(status: %w[completed skipped]) }
+  scope :incomplete, -> { where.not(status: COMPLETED_STATES) }
+  scope :complete, -> { where(status: COMPLETED_STATES) }
```
peetucket, line 18: Some refactoring here to reduce duplication, and because I added a new helper method to ask if the step is completed.

PR #418, opened 2020-09-18
Set completed_at date but never reset it

mjgiarlo:
Fixes sul-dlss/argo#2252

## Why was this change made?

To avoid re-setting completed_at for past completed steps, e.g., when the WorkflowCreator sweeps through and marks old version's steps as not the active version.

CI, and will test in stage.

None

### DIFF
--- a/app/models/workflow_step.rb
+++ b/app/models/workflow_step.rb
@@ -11,7 +11,7 @@ class WorkflowStep < ApplicationRecord
   validate  :workflow_exists, on: :create
   validate  :process_exists_for_workflow, on: :create
 
-  before_save :set_completed_at, if: :completed?
+  before_save :maybe_set_completed
 
   scope :lifecycle, -> { where.not(lifecycle: nil) }
   scope :incomplete, -> { where.not(status: COMPLETED_STATES) }
@@ -34,16 +34,19 @@ class WorkflowStep < ApplicationRecord
                   version: version)
   end
 
-  # callback to set the completed_at column to the current time if we are completing a step
-  def set_completed_at
-    self.completed_at = Time.now
+  # callback to set the completed_at column to the current time if we are
+  # completing a step. if completed_at is already set, leave the value untouched.
+  def maybe_set_completed
+    return unless completed?
+
+    self.completed_at ||= Time.now
   end
 
   ##
   # indicate if this step is marked as completed
   # @return [boolean]
   def completed?
-    COMPLETED_STATES.include? status
+    COMPLETED_STATES.include?(status)
   end
 
   ##
--- a/spec/models/workflow_step_spec.rb
+++ b/spec/models/workflow_step_spec.rb
@@ -136,11 +136,33 @@ RSpec.describe WorkflowStep do
 
   describe '#completed?' do
     it 'indicates if the step is not completed' do
-      expect(step.completed?).to be_falsey
+      expect(step).not_to be_completed
     end
 
     it 'indicates if the step is completed' do
-      expect(completed_step.completed?).to be_truthy
+      expect(completed_step).to be_completed
+    end
+  end
+
+  describe '#save' do
+    context 'when completed and not already completed' do
+      it 'sets a value for completed_at' do
+        expect(step.completed_at).to be_nil
+        step.status = 'completed'
+        step.save
+        expect(step.completed_at.to_i).to eq(step.updated_at.to_i) # use .to_i to ignore millisecond comparison
+      end
+    end
+
+    context 'when completed and already completed' do
+      let(:completed_at) { completed_step.completed_at }
+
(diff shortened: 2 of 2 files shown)


peetucket (approved):
I guess it makes sense that it can keep completing because we store the date as completed_at (and not just a boolean).  If it was a boolean, activerecord would realize nothing actually changed and wouldn't bother saving the record again.

Looks fine, awkward method name (`completed_and_not_already_completed`) but it is what it is.

app/models/workflow_step.rb
```diff
@@ -39,11 +40,21 @@ def set_completed_at
...
+  def completed_and_not_already_completed?
+    completed? && !already_completed?
+  end
+
   ##
   # indicate if this step is marked as completed
   # @return [boolean]
   def completed?
-    COMPLETED_STATES.include? status
+    COMPLETED_STATES.include?(status)
```
peetucket, line 52: any reason parens were added?  is there a preferred rubocop style?

mjgiarlo: Ah, because I was originally going to add a condition to this method and didn't want the parser to get confused, but then I backed it out and left the parens.

mjgiarlo:
@peetucket :speech_balloon: 
> I guess it makes sense that it can keep completing because we store the date as completed_at (and not just a boolean). If it was a boolean, activerecord would realize nothing actually changed and wouldn't bother saving the record again.

Yeppers.
 
> Looks fine, awkward method name (`completed_and_not_already_completed`) but it is what it is.

Open to suggestions! I was similarly :unamused: with that name but the others I came up with were worse in my judgment.

app/models/workflow_step.rb
```diff
@@ -11,7 +11,8 @@ class WorkflowStep < ApplicationRecord
   validate  :workflow_exists, on: :create
   validate  :process_exists_for_workflow, on: :create
 
-  before_save :set_completed_at, if: :completed?
+  # Only set the completed_at field if status is completed and completed_at is unset
+  before_save :set_completed_at, if: :completed_and_not_already_completed?
```
jcoyne, line 15: How about calling `before_save :maybe_set_completed`

app/models/workflow_step.rb
```diff
@@ -39,11 +40,21 @@ def set_completed_at
...
   ##
   # indicate if this step is marked as completed
   # @return [boolean]
   def completed?
-    COMPLETED_STATES.include? status
+    COMPLETED_STATES.include?(status)
+  end
+
+  # Check if there is already a completed_at date
+  def already_completed?
```
jcoyne, line 56: and then this could be `maybe_set_completed` which would have all the conditions. (e.g. `completed? && !completed_at`)

mjgiarlo: ok, I like it. will do.

app/models/workflow_step.rb
peetucket: `before_save :set_completed_at, :if !already_completed?`  ?

mjgiarlo:
@jcoyne have another look?

app/models/workflow_step.rb
peetucket: can you do that ?  do we even need `completed_and_not_already_completed?`

peetucket: or `:set_completed_at, :unless already_completed?`

app/models/workflow_step.rb
```diff
@@ -34,16 +34,25 @@ def as_milestone(xml)
                   version: version)
   end
 
-  # callback to set the completed_at column to the current time if we are completing a step
-  def set_completed_at
+  # callback to set the completed_at column to the current time if we are
+  # completing a step and completed_at is not already set
+  def maybe_set_completed
+    return unless completed?
+    return if already_completed?
```
jcoyne, line 41: Can you do this as `self.completed_at ||= Time.now` ?

app/models/workflow_step.rb
```diff
@@ -34,16 +34,25 @@ def as_milestone(xml)
...
   ##
   # indicate if this step is marked as completed
   # @return [boolean]
   def completed?
-    COMPLETED_STATES.include? status
+    COMPLETED_STATES.include?(status)
+  end
+
+  # Check if there is already a completed_at date
+  def already_completed?
```
jcoyne, line 54: I think the fact that there is a `completed?` method and an `already_completed?` method is confusing.

app/models/workflow_step.rb
mjgiarlo: done

app/models/workflow_step.rb
mjgiarlo: agree, change made

jcoyne (approved):

PR #442, opened 2021-01-12
Connect to rabbitmq

jcoyne:
## Why was this change made?

For testing https://github.com/sul-dlss/operations-tasks/issues/2206
For exploring https://github.com/sul-dlss/happy-heron/issues/797

Working from this blog https://www.monterail.com/blog/2014/event-sourcing-on-rails-with-rabbitmq

### DIFF
--- a/app/controllers/workflows_controller.rb
+++ b/app/controllers/workflows_controller.rb
@@ -79,7 +79,6 @@ class WorkflowsController < ApplicationController
         version: params[:version]
       )
     ).create_workflow_steps
-    SendUpdateMessage.publish(druid: params[:druid])
   end
 
   private
--- a/app/services/send_rabbitmq_message.rb
+++ b/app/services/send_rabbitmq_message.rb
@@ -17,7 +17,9 @@ class SendRabbitmqMessage
   # put any specific RabbitMQ settings
   # like host or port
   def self.connection
-    @connection ||= Bunny.new(hostname: Settings.rabbitmq.hostname).tap(&:start)
+      @connection ||= Bunny.new(hostname: Settings.rabbitmq.hostname,
+                                username: Settings.rabbitmq.username,
+                                password: Settings.rabbitmq.password).tap(&:start)
   end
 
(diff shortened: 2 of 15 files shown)


app/controllers/workflows_controller.rb
```diff
@@ -79,7 +79,6 @@ def create
         version: params[:version]
       )
     ).create_workflow_steps
-    SendUpdateMessage.publish(druid: params[:druid])
```
mjgiarlo, line 82: This moved to the WorkflowCreator? Can you say more about this change?

app/services/send_rabbitmq_message.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# Send a message to a RabbitMQ exchange that an item has been updated.
+class SendRabbitmqMessage
```
mjgiarlo, line 4: This class is showing as entirely uncovered by tests. Are you thinking it doesn't need unit tests?

jcoyne: Any test we could do here is merely going to be a mirror image of the code but in mocks.  I don't particularly thing that that sort of testing is useful.  Furthermore, we've tested this code already in qa and stage.

app/controllers/workflows_controller.rb
jcoyne: I believe this to be unnecessary, because `WorkflowCreator#create_workflow_steps` calls `WorkflowCreator#enqueue` which calls `SendUpdateMessage.publish`

mjgiarlo: OK.

app/services/send_rabbitmq_message.rb
mjgiarlo: Yeah, mocking the channel topic to test if it receives `#publish` with the args we expect does not seem worth the machinery or time.

mjgiarlo (approved):

PR #470, opened 2021-06-15
add validate-moab step to preservationIngestWF definition

### LINKED ISSUE #463: update preservationIngestWF to include validate-moab step, make sure it works
blocked by https://github.com/sul-dlss/preservation_robots/issues/315

The new preservationIngestWF step, `validate-moab` , should fall between the `update-moab` and `update-catalog` steps.  It must succeed for `update-catalog` to start.  It *may* only be a matter of adding the new step to this file:  https://github.com/sul-dlss/workflow-server-rails/blob/main/config/workflows/preservationIngestWF.xml  and updating the sequence numbers, and then deploying the workflow server.   

Step name:  validate-moab
Label:  "Verify the Moab on local disk passes validation, including checksums"

```xml
<workflow-def id="preservationIngestWF" repository="sdr">
  <process name="start-ingest" sequence="1" status="completed">
    <label>Initiate preservation ingest workflow</label>
  </process>
...
  <process name="update-moab" sequence="4">
    <label>Create/update Moab object from deposit bag</label>
    <prereq>validate-bag</prereq>
  </process>
INSERT NEW STEP HERE
  <process name="update-catalog" sequence="5">
    <label>Create/update Preservation Catalog entry</label>
    <prereq>update-moab</prereq>
  </process>
  <process name="complete-ingest" sequence="6">
    <label>Clean up workspace; transfer control back to accessioning</label>
    <prereq>update-catalog</prereq>
  </process>
</workflow-def>
```

Obviously, this must be tested thoroughly in stage and/or qa.   You should also check the information in Argo to be sure the step ran correctly.   You should work with Andrew to confirm a bad Moab errors appropriately in the workflow and stops accessioning. 

### Gotcha

- changing an active workflow REQUIRES updates to the workflow of any objects that have active, INCOMPLETE workflows, on all deployed environments (qa, stage, prod)  For the preservationIngestWF, this is a very small number of objects in production.  (You will also need to do this for objects in -stage and -qa)
   - instructions for how to do this:  https://github.com/sul-dlss/workflow-server-rails/wiki/How-to-add-missing-workflow-steps  (obviously with different workflow and step names)
   - talk to Andrew re:  which objects are affected in production.  You can also tell from this facet in Argo:

image

or from the workflow-grid at https://argo.stanford.edu/report/workflow_grid

---------------
Background:  

This should be the final piece of ensuring that anything that doesn't pass full checksum validation _during_ `preservationIngestWF`, before preservation_catalog is officially updated, and before `preservationIngestWF` can hand back to `accessionWF` halts the accessioning process for that item, and thus prevents automatic cleanup of deposited content.

closes sul-dlss/preservation_catalog#1691

peetucket:
## Why was this change made?

Fixes #463 - add validate-moab step

Hold until we test on qa/stage 

closes sul-dlss/preservation_catalog#1691

### DIFF
--- a/config/workflows/preservationIngestWF.xml
+++ b/config/workflows/preservationIngestWF.xml
@@ -15,11 +15,15 @@
     <label>Create/update Moab object from deposit bag</label>
     <prereq>validate-bag</prereq>
   </process>
-  <process name="update-catalog" sequence="5">
-    <label>Create/update Preservation Catalog entry</label>
+  <process name="validate-moab" sequence="5">
+    <label>Verify the Moab on local disk passes validation, including checksums</label>
     <prereq>update-moab</prereq>
   </process>
-  <process name="complete-ingest" sequence="6">
+  <process name="update-catalog" sequence="6">
+    <label>Create/update Preservation Catalog entry</label>
+    <prereq>validate-moab</prereq>
+  </process>
+  <process name="complete-ingest" sequence="7">
     <label>Clean up workspace; transfer control back to accessioning</label>
     <prereq>update-catalog</prereq>
   </process>


jcoyne (approved):

ndushay (approved):

ndushay:
I was able to affirm that the step is hooked up and will execute https://github.com/sul-dlss/preservation_catalog/pull/1725 is merged and deployed.

However, we still want Andrew to test problematic objects

ndushay:
Also, we'll need this PR for prod:  https://github.com/sul-dlss/shared_configs/pull/1667

jmartin-sul:
_Converted to draft to prevent premature merge._

Tested in stage using infra integration test suite, and the worker re-allocation ([see here](https://github.com/sul-dlss/shared_configs/pull/1670)) appears to be fine, but I did discover a queue locking issue in pres cat that prevents a job from being run twice with the same parameters (more or less) in rapid succession (because the job doesn't clean up the resque-lock entry it leaves as would be expected). The upshot is that an object can't be accessioned twice in rapid succession (as happens in `create_preassembly_image_spec.rb` in the integration suite) till that queue locking bug is fixed.

I believe I have a fix for the lock cleanup issue, in pres cat, but haven't had a chance to test it yet.

I believe this PR will likely remain unchanged, as the rest of the mechanism works fine.  If I clean up the job manually and retry queuing the pres cat job that pres robots tried to queue for the new `validte-moab` step, the step completes successfully and the version finishes accessioning as expected (without doing that, the dropped attempt to queue the job in pres cat results in this new step hanging forever).

jmartin-sul:
fwiw, this pres cat issue has likely been there with all jobs all along, this is just the first time it's been more than a curiosity.  e.g. in the past, i've noticed that i couldn't queue the same checksum validation job twice in rapid succession.

jmartin-sul:
@andrewjbtw re-tested with this branch deployed, and with both https://github.com/sul-dlss/preservation_catalog/pull/1729 and https://github.com/sul-dlss/preservation_catalog/pull/1728 both having been merged.

for peace of mind over the weekend, we'll merge this and deploy monday, separately from dep updates.

PR #479, opened 2021-07-14
Add update doi step to accessionWF

### LINKED ISSUE #478: Create a new workflow step in accessionWF
This should be after shelve and publish.

Ref https://github.com/sul-dlss/dor-services-app/issues/2912
Ref https://github.com/sul-dlss/common-accessioning/issues/780

jcoyne:
## Why was this change made?

Fixes #478

### DIFF
--- a/config/workflows/accessionWF.xml
+++ b/config/workflows/accessionWF.xml
@@ -31,25 +31,30 @@
     <label>Publish Metadata</label>
     <prereq>shelve</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" name="provenance-metadata" sequence="9">
+  <process batch-limit="100" error-limit="10" name="update-doi" sequence="9">
+    <label>Update DOI Metadata</label>
+    <prereq>publish</prereq>
+  </process>
+  <process batch-limit="1" error-limit="5" name="provenance-metadata" sequence="10">
     <label>Provenance Metadata</label>
     <prereq>publish</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" name="sdr-ingest-transfer" sequence="10">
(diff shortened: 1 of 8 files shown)


config/workflows/accessionWF.xml
```diff
@@ -31,25 +31,30 @@
     <label>Publish Metadata</label>
     <prereq>shelve</prereq>
   </process>
-  <process batch-limit="1" error-limit="5" name="provenance-metadata" sequence="9">
+  <process batch-limit="100" error-limit="10" name="update-doi" sequence="9">
```
ndushay, line 34: i have no idea what batch-limit does, but I expect that you have deliberately set it to 100 for this step and left it at 1 for the subsequent steps.

spec/requests/version_close_spec.rb
```diff
@@ -12,6 +12,6 @@
   it 'closes the version' do
     post "/objects/#{druid}/versionClose?version=3"
     expect(response).to be_successful
-    expect(WorkflowStep.where(druid: druid).count).to eq 13
+    expect(WorkflowStep.where(druid: druid).count).to eq 14
```
ndushay, line 15: seems like it would be easier to have a constant instead of updating this in so many places, but 🤷🏻‍♀️

ndushay (approved):

config/workflows/accessionWF.xml
jcoyne: I don't think it does anything, I just copied from a different step that had 100.

spec/requests/version_close_spec.rb
jcoyne: This happens so infrequently that I don't mind the extra 5 minutes.

PR #510, opened 2021-12-09
README: hydrus becomes happy-heron

ndushay:
## Why was this change made?

bring README more up to date

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Maintainability](https://api.codeclimate.com/v1/badges/cc78d20264a4eaf8a782/maintainability)](https://codeclimate.com/github/sul-dlss/workflow-server-rails/maintainability)
 [![](https://images.microbadger.com/badges/image/suldlss/workflow-server.svg)](https://microbadger.com/images/suldlss/workflow-server "Get your own image badge on microbadger.com")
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,7 @@
 [![Maintainability](https://api.codeclimate.com/v1/badges/cc78d20264a4eaf8a782/maintainability)](https://codeclimate.com/github/sul-dlss/workflow-server-rails/maintainability)
 [![](https://images.microbadger.com/badges/image/suldlss/workflow-server.svg)](https://microbadger.com/images/suldlss/workflow-server "Get your own image badge on microbadger.com")
 
-This is a Rails-based workflow service that replaced SDR's Java-based workflow service.  It is consumed by the users of dor-workflow-client (argo, hydrus, hydra_etd, pre-assembly, dor-indexing-app, robots) and *soon* the goobi application (currently proxying through dor-services-app).
+This is a Rails-based workflow service that replaced SDR's Java-based workflow service.  It is consumed by the users of dor-workflow-client (argo, happy-heron (h2), hydra_etd, pre-assembly, dor-indexing-app, robots) and *soon* the goobi application (currently proxying through dor-services-app).
```
jcoyne, line 8: Hydrus was actually using the workflow service, but h2 doesn't. I think we can just remove this from the list.

jcoyne: We ought to add dor-services-app though.

jcoyne: Here's a good list to draw from: https://github.com/sul-dlss/dor-workflow-client/network/dependents

ndushay: I'll kill this PR and do it over.

PR #511, opened 2021-12-09
README: update lists of direct consumers

ndushay:
## Why was this change made?

Out of date documentation is confusing.

based on dependency diagram from workflow-client gem with gem called out in Gemfile.


jcoyne (approved):
This is a good change. I think hydra_etd can come off the list too, but not a blocker.

jcoyne (commented):
It would be really good to get dor-services-app on there too.

ndushay:
@jcoyne feel free to update again;  I'm merging this as a step forward.

ndushay:
> This is a good change. I think hydra_etd can come off the list too, but not a blocker.

See https://github.com/sul-dlss/hydra_etd/blob/main/Gemfile#L14

jcoyne:
Huh. I thought we moved all that into the database.  Thanks for checking.

PR #517, opened 2022-01-05
Update listen gem for Ruby 3.0

lwrubel:
## Why was this change made?
Ruby 3.0 techstack upgrades

CircleCI and deployed to qa VM with ruby-3.0.3

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,7 @@ group :development do
   gem 'capistrano-passenger', require: false
   gem 'capistrano-rails', require: false
   gem 'dlss-capistrano', require: false
-  gem 'listen'
+  gem 'listen', '~> 3.7'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -35,7 +35,7 @@ group :development do
   gem 'capistrano-passenger', require: false
   gem 'capistrano-rails', require: false
   gem 'dlss-capistrano', require: false
-  gem 'listen', '>= 3.0.5', '< 3.2'
+  gem 'listen'
```
jcoyne, line 38: Typically we want to leave a version spec in Gemfile if possible.  In this case '~> 3.7' would be a good choice.  This allows the minor version to float up but it won't automatically upgrade to 4.0 without changing the version spec.

PR #520, opened 2022-01-06
Updating CircleCI to Ruby to 3.0.3 and listen gem

lwrubel:
## Why was this change made?

Ruby 3.0 techstack upgrades

CircleCI and deployed to qa and stage VM with ruby-3.0.3

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
 jobs:
   test:
     docker:
-    - image: circleci/ruby:2.7.1-node
+    - image: cimg/ruby:3.0.3-node
       environment:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -4,11 +4,11 @@ executors:
     environment:
       IMAGE_NAME: suldlss/workflow-server
     docker:
-    - image: circleci/buildpack-deps:stretch
+    - image: cimg/buildpack-deps:stretch
```
jcoyne, line 7: I don't think this image exists.

lwrubel: You're right, thanks. Will revert that for now.

PR #565, opened 2022-06-08
Add new CDXJ steps to wasCrawlDisseminationWF

jcoyne:
Ref https://github.com/sul-dlss/was_robot_suite/issues/449

CI


jcoyne:
@lwrubel I didn't see any requirements about where to put these, so I just put them after the existing corresponding steps.

jcoyne:
@lwrubel fixed. I missed the requirement about putting them after path-indexer.

lwrubel (approved):

PR #571, opened 2022-06-22
remove skipped processes in gisAssemblyWF

### LINKED ISSUE #567: Remove skipped steps from the gisAssemblyWF
https://github.com/sul-dlss/workflow-server-rails/blob/main/config/workflows/gisAssemblyWF.xml#L78-L85

peetucket:

Fixes #567 - remove steps from gisAssemblyWF template that are always skipped

Existing tests

### DIFF
--- a/config/workflows/gisAssemblyWF.xml
+++ b/config/workflows/gisAssemblyWF.xml
@@ -75,12 +75,4 @@
     <prereq>load-geo-metadata</prereq>
     <label>Finalize assembly workflow to prepare for assembly/delivery/discovery</label>
   </process>
-  <process name="start-assembly-workflow" sequence="20" status="skipped">
-    <prereq>finish-gis-assembly-workflow</prereq>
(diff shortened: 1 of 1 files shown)


ndushay:
Unfortunately, this also requires some footwork to ensure there are no objects with gisAssemblyWF in flight that will try to execute the removed steps ... 

I believe there is a way to check for these cases in Argo ... and I vaguely remember Andrew looking for something like this ... it's unfortunate it didn't make it into JCoynes issue, but it should be done as part of the work for #571

Here is an old breadcrumb about it:  https://github.com/sul-dlss/workflow-server-rails/wiki/How-to-add-missing-workflow-steps   (like, how you work with those objects with in-flight workflow to get the workflow to finish for that object ...)

ndushay:
Huh - based on the below from Argo, this is going to be fine.  Merging.

image

ndushay (approved):

PR #582, opened 2022-07-13
update README and specs to reflect absence of repo parameter

ndushay:

since this code no longer takes the repo parameter, no need for it to be called out in specs or in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,26 +76,24 @@ To shut down postgres afterwards,
 - ```docker-compose down``` afterwards
 
 ## Routes:
-`GET    /:repo/objects/:druid/lifecycle` - Returns the milestones in the lifecycle that have been completed
-
 `GET    /objects/:druid/lifecycle` - Returns the milestones in the lifecycle that have been completed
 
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -76,26 +76,24 @@ To shut down postgres afterwards,
 - ```docker-compose down``` afterwards
 
 ## Routes:
-`GET    /:repo/objects/:druid/lifecycle` - Returns the milestones in the lifecycle that have been completed
-
```
ndushay, line 80: routes with repo in them were removed a month ago.

spec/requests/workflows/single_workflow_for_object_spec.rb
```diff
@@ -3,77 +3,38 @@
 require 'rails_helper'
 
 RSpec.describe 'Get a single workflow for an object', type: :request do
-  context 'when repo is supplied (deprecated)' do
```
ndushay, line 6: AFAICT, there was no different in the specs inside the two contexts (with vs. without the repo param)

jcoyne (approved):

PR #630, opened 2023-02-07
restore the behavior of returning a 404 when trying to update a missing workflow step

jmartin-sul:

discussed after standup just now.  consensus was to change this back to the way things were before https://github.com/sul-dlss/workflow-server-rails/pull/588 because

* this seems correct semantically (a `PUT` to a missing resource should be a 404, not a 500).
* and practically, preservation_catalog depends on the 404 (`Dor::MissingWorkflowException` from dor-workflow-client) to create `preservationAuditWF` when it isn't already present (see e.g. https://github.com/sul-dlss/preservation_catalog/blob/2b4d75030b673fe6f2a1c65e87a54febf851ce59/app/services/audit_reporters/audit_workflow_reporter.rb#L72-L100).

links for context:
* pres cat HB alert about being unable to PUT against preservationAuditWF for the druid: https://app.honeybadger.io/projects/54415/faults/93228428
* pres robots HB alert for the 500 it gets back from pres cat: https://app.honeybadger.io/projects/55564/faults/93065492
* example HB alert for WFS 500 instead of 404: https://app.honeybadger.io/projects/58890/faults/93131970
* druid i was testing with this past friday: https://argo.stanford.edu/view/druid:cg595vp1970

unit tests

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -16,10 +16,15 @@ class StepsController < ApplicationController
 
   # Update a single WorkflowStep
   # If there are next steps, they are enqueued.
-  def update # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/MethodLength
+  def update
     parser = ProcessParser.new(process_from_request_body, use_default_lane_id: false)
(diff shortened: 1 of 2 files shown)


edsu (approved):
I noticed a lint issue, but otherwise I agree that the 404 makes more sense here than a 500.

justinlittman (approved):

PR #649, opened 2023-05-11
Upgrade to Ruby 3.2.2

mjgiarlo:

Keep pace with Ruby releases

CI and will be int-tested in stage


jmartin-sul (approved):
i do see a couple uncovered lines (according to code climate plugin), but not worried about them, ✅ 

https://github.com/sul-dlss/workflow-server-rails/pull/649/files#diff-d20dd009c8a404c9c7abba11f22b6138b9f6d6edc86908f7e924ea23b918c38eR8
https://github.com/sul-dlss/workflow-server-rails/pull/649/files#diff-bdc6354e8ad76ee945d20702dbcb9b52dd64c8fb1a95d8228210bc7f71895d0eR7

PR #652, opened 2023-05-16
Adjust workflow monitoring periodicity and trigger timestamps

mjgiarlo:

We discussed making this change in our team dev planning meeting and ran it by the product owner, Andrew, who says running the job every 6 hours and only flagging things 48-hours old is sufficient.

CI + deployed to QA to verify crontab change


config/schedule.rb
```diff
@@ -27,7 +27,7 @@
 
 # NOTE: this job uses HB checkins.
 # If you change the schedule, edit the HB checkin at https://app.honeybadger.io/projects/58890/check_ins
```
mjgiarlo, line 29: I already did ☝🏻

jmartin-sul (approved):

PR #666, opened 2023-07-17
Adds update orcid work step to accession workflow.

### LINKED ISSUE #664: Add create/update orcid work step to accessioning workflow
It should be after update-datacite step and naming should align with https://github.com/sul-dlss/common-accessioning/issues/1073.

justinlittman:
closes #664

QA

### DIFF
--- a/config/workflows/accessionWF.xml
+++ b/config/workflows/accessionWF.xml
@@ -27,6 +27,10 @@
     <label>Update DOI Metadata</label>
     <prereq>publish</prereq>
   </process>
+  <process batch-limit="100" error-limit="10" name="update-orcid-work">
+    <label>Update ORCID work</label>
+    <prereq>publish</prereq>
+  </process>
   <process batch-limit="1" error-limit="5" name="sdr-ingest-transfer">
     <label>Initiate Ingest into Preservation</label>
     <prereq>publish</prereq>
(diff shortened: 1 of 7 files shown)


config/workflows/accessionWF.xml
```diff
@@ -27,6 +27,10 @@
     <label>Update DOI Metadata</label>
     <prereq>publish</prereq>
   </process>
+  <process batch-limit="100" error-limit="10" name="update-orcid-work">
+    <label>Update ORCID work</label>
+    <prereq>publish</prereq>
+  </process>
```
mjgiarlo, line 33: @justinlittman I see this process has been set up similarly to `update-doi` in terms of the configured limits and the prerequisite which makes me wonder if `update-orcid-work` should also be a prerequisite of `end-accession` below (as is `update-doi`).

db/schema.rb
```diff
@@ -28,10 +27,10 @@
     t.integer "version"
     t.text "note"
     t.string "lane_id", default: "default", null: false
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at", precision: nil, null: false
+    t.datetime "updated_at", precision: nil, null: false
     t.boolean "active_version", default: false
-    t.datetime "completed_at"
+    t.datetime "completed_at", precision: nil
```
mjgiarlo, line 33: Do we want/need the changes to the DB schema in this PR?

config/workflows/accessionWF.xml
justinlittman: It shouldn't matter, but I added anyway.

lwrubel (approved):

PR #707, opened 2024-01-18
reset-geowebcache is never being run

peetucket:

Fixes https://github.com/sul-dlss/gis-robot-suite/issues/401

This step is never being run, because it appears to be misconfigured in the workflow definition (nothing has it as a pre-req...and it is the pre-req of itself).  This step was added way back here (3.5 years ago) incorrectly: https://github.com/sul-dlss/workflow-server-rails/commit/5f495a3444f764532a0ebadf4388497eafa9fee1 so it looks like it has never been run.

Not sure if this is a problem or not.

I manually ran the logic for the robot on stage for this object druid:st417qr9346 and it completed successfully (though it's not clear to me how to verify it did what was intended).  https://argo-stage.stanford.edu/view/druid:st417qr9346

### DIFF
--- a/config/workflows/gisDeliveryWF.xml
+++ b/config/workflows/gisDeliveryWF.xml
@@ -16,11 +16,11 @@
     <label>Load layers into GeoServer</label>
   </process>
   <process name="reset-geowebcache">
-    <prereq>reset-geowebcache</prereq>
+    <prereq>load-geoserver</prereq>
     <label>Reset GeoWebCache for the layer</label>
   </process>
   <process name="finish-gis-delivery-workflow">
-    <prereq>load-geoserver</prereq>
+    <prereq>reset-geowebcache</prereq>
     <label>Finalize delivery workflow for the object</label>
   </process>
 </workflow-def>


peetucket:
Run a druid manually (e.g. on kurma-robots-stage-0) via:

```
ROBOT_ENVIRONMENT=staging bin/console
druid='druid:st417qr9346'
workflow_service = LyberCore::WorkflowClientFactory.build(logger: nil)
workflow_service.update_status(druid:, workflow: 'gisDeliveryWF', process: 'reset-geowebcache', status: 'queued') 
robot = Robots::DorRepo::GisDelivery::ResetGeowebcache.new
robot.perform druid
```

jmartin-sul (commented):
this seems like a sensible change, but i think i'd want to understand how to verify that the step does what's intended; i'm not sure how to do that either.

on top of that, i couldn't resist a little querying to see whether there'd been any real world usage, and it seems like there has, starting a few weeks after the commit you referenced that adds this step.  query result from prod:

```
$ bin/rails c -e p
Loading production environment (Rails 7.0.8)
3.2.0 :001 > WorkflowStep.where(process: 'reset-geowebcache').pluck('MIN(created_at)')
 => [2020-07-29 16:20:26.092129 UTC] 
3.2.0 :002 > WorkflowStep.where(process: 'reset-geowebcache').count
 => 6564 
```

i sort of expected a non-zero number of rows, because maybe there was some manual testing like what you did for this PR.  but 6.5k+ is like a lot more than i'd expect.

i'd checked because i had this vague memory that prereqs weren't actually a gate on workflow steps running; but based on a look at what i think is the relevant code, i think i was mis-remembering, and prereqs do indeed seem to be meaningful: https://github.com/sul-dlss/workflow-server-rails/blob/e8da8431520ddf89977d515a7e913b4b0129c922/app/services/next_step_service.rb#L35

peetucket:
If we accept this change, we should export all druids with this step in waiting (see https://argo.stanford.edu/report?f%5Bcontent_type_ssim%5D%5B%5D=geo&f%5Bwf_wps_ssim%5D%5B%5D=gisDeliveryWF%3Areset-geowebcache%3Awaiting), then write a script to run on the gis robot server (see https://github.com/sul-dlss/workflow-server-rails/pull/707#issuecomment-1897582478) to ensure all previous druids have had this step run.

jmartin-sul (approved):
did a bit more querying, and discussed w/ peter after standup:
* there are 400 something `completed` instances of this step in prod, dating from august 2020 to now, but peter said that kim said that she's manually marked this step as completed in the past when the noise in the workflow grid from all the un-run `waiting` steps gets in the way too much.  and peter and i are now both pretty sure from our read of the workflow code that the step has been applied but not automatically run since it was added to the workflow definition in july 2020.  there are 6000+ `waiting` entries on prod, which is also congruent with this read of things.
* i'm going to mark this approved, and peter's going to deploy to stage and do some before/after testing (i.e. run integration tests before merging this, see the step get applied but left un-run in `waiting`, merge this, deploy it, run integration tests again, hopefully see the step run automatically).

PR #715, opened 2024-01-31
Add reload-geoserver to gisDeliveryWF

aaron-collier:

Part of: https://github.com/sul-dlss/gis-robot-suite/pull/748 (DRAFT)

### DIFF
--- a/config/workflows/gisDeliveryWF.xml
+++ b/config/workflows/gisDeliveryWF.xml
@@ -23,8 +23,12 @@
     <prereq>reset-geowebcache</prereq>
     <label>Finalize delivery workflow for the object</label>
   </process>
-  <process name="start-accession-workflow">
+  <process name="reload-geoserver">
     <prereq>finish-gis-delivery-workflow</prereq>
+    <label>Automatically reload the geo server</label>
+  </process>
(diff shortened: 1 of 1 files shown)


config/workflows/gisDeliveryWF.xml
```diff
@@ -23,6 +23,10 @@
     <prereq>reset-geowebcache</prereq>
     <label>Finalize delivery workflow for the object</label>
   </process>
+  <process name="reload-geoserver">
+    <prereq>finish-gis-delivery-workflow</prereq>
+    <label>Automatically reload the geo server</label>
+  </process>
   <process name="start-accession-workflow">
     <prereq>finish-gis-delivery-workflow</prereq>
```
ndushay, line 31: do you want this step to be sequential after reload-geoserver?    I'm not sure if we have non-sequential workflows.  Maybe.

ndushay (commented):
a comment for your consideration.

config/workflows/gisDeliveryWF.xml
aaron-collier: Good catch, thank you.

ndushay (approved):

PR #720, opened 2024-02-07
Rename generate-content-metadata step in workflow

### LINKED ISSUE #719: rename generate-content-metadata robot step to generate-structural-metadata
The current name made sense when the robot was generating `contentMetadata.xml` for Fedora (or for later translation to Cocina after we removed Fedora from SDR).  Now that the step is directly producing Cocina structural metadata, we should rename the step accordingly.

See https://github.com/sul-dlss/workflow-server-rails/blob/3808b606328440372d1dd71386118eb8c9eaec6c/config/workflows/gisAssemblyWF.xml#L46-L53

Before deploying, either make sure that nothing in the deploy env (e.g stage or prod) is in process in `gisAssemblyWF`, or stop the GIS robots and migrate the active workflow steps in workflow-server-rails' database to use the new name.  Some [useful related info is here](https://github.com/sul-dlss/workflow-server-rails/wiki/What-happens-when-a-step-is-added-or-removed-from-a-workflow-definition), although that wiki entry is about adding/removing workflow steps, not renaming, so not exactly the same.

This should be coordinate with https://github.com/sul-dlss/gis-robot-suite/issues/784 (whoever picks up one should likely pick up the other at the same time also)

peetucket:

Fixes #719 - rename generate-content-metadata to generate-structural

Must go with https://github.com/sul-dlss/gis-robot-suite/pull/787

HOLD for coordination with gis-robot-suite change

CI

### DIFF
--- a/config/workflows/gisAssemblyWF.xml
+++ b/config/workflows/gisAssemblyWF.xml
@@ -48,7 +48,7 @@
     <label>Generate cocina structural</label>
   </process>
   <process name="finish-gis-assembly-workflow">
-    <prereq>generate-content-metadata</prereq>
+    <prereq>generate-structural</prereq>
     <label>Finalize assembly workflow to prepare for assembly/delivery/discovery</label>
   </process>
   <process name="start-delivery-workflow">
--- a/config/workflows/gisAssemblyWF.xml
+++ b/config/workflows/gisAssemblyWF.xml
@@ -43,9 +43,9 @@
     <prereq>normalize-data</prereq>
     <label>Extract bounding box from data for MODS record</label>
   </process>
-  <process name="generate-content-metadata">
+  <process name="generate-structural">
     <prereq>extract-boundingbox</prereq>
-    <label>Generate contentMetadata manifest</label>
+    <label>Generate cocina structural</label>
   </process>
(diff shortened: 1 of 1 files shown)


config/workflows/gisAssemblyWF.xml
```diff
@@ -43,9 +43,9 @@
...
     <label>Extract bounding box from data for MODS record</label>
   </process>
-  <process name="generate-content-metadata">
+  <process name="generate-structural">
     <prereq>extract-boundingbox</prereq>
-    <label>Generate contentMetadata manifest</label>
+    <label>Generate cocina structural</label>
   </process>
   <process name="finish-gis-assembly-workflow">
     <prereq>generate-content-metadata</prereq>
```
lwrubel, line 51: I think this should be `generate-structural`?

lwrubel, line 51: I think this should be `generate-structural`?

peetucket: Good catch, fixed.

lwrubel (approved):

jmartin-sul (approved):

PR #737, opened 2024-03-26
Remove content-metadata and desc-metadata steps from accessionWF

### LINKED ISSUE #736: Remove descriptive-metadata and content-metadata steps from accessioning WF
They are no longer necessary since no assembly workflows create XML files.

jcoyne:
Fixes #736

### DIFF
--- a/config/workflows/accessionWF.xml
+++ b/config/workflows/accessionWF.xml
@@ -3,21 +3,13 @@
   <process lifecycle="submitted" name="start-accession" status="completed">
     <label>Start Accessioning</label>
   </process>
-  <process batch-limit="1000" error-limit="10" lifecycle="described" name="descriptive-metadata">
-    <label>Any descMetadata.xml is converted and posted to cocina object; no-op if no such file exists (cocina object may already have information)</label>
-    <prereq>start-accession</prereq>
-  </process>
-  <process batch-limit="1000" error-limit="10" name="content-metadata">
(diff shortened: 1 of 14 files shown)


config/workflows/accessionWF.xml
```diff
@@ -3,21 +3,13 @@
...
-  </process>
   <process batch-limit="1000" error-limit="10" name="technical-metadata">
     <label>Creates the technical metadata by calling technical-metadata-service, only for item objects with files</label>
-    <prereq>content-metadata</prereq>
+    <prereq>start-accession</prereq>
   </process>
   <process batch-limit="1" error-limit="5" name="shelve">
     <label>Shelve content in Digital Stacks</label>
-    <prereq>technical-metadata</prereq>
+    <prereq>start-accession</prereq>
```
justinlittman, line 12: Can you say more about why you made this change?

jcoyne: There was an inaccuracy here. There is no prerequisite on technical metadata. This can start as soon as the workflow is initiated.

andrewjbtw: I think technical metadata had to wait for the final content metadata step in case any files were added to the list at that point. If content metadata (structural metadata) now comes into the workflow fully-formed then this change should be ok.

andrewjbtw (commented):
Github once again led me to start a review when I just wanted to post a comment.

justinlittman (approved):

PR #745, opened 2024-04-03
Add version metadata to objects

### LINKED ISSUE #743: Create migration for workflow variables
We will need to pass information into a workflow when creating it.  This will require new workflow variables in the data model.

### LINKED ISSUE #744: Update workflow API to handle workflow variables
Workflow variables are available to consumers of the workflow service, and thus the API needs to be able to handle variables being sent in and retrieved

peetucket:

Fixes #743 and #744 

Borrows ideas from #746 

sul-dlss/dor-workflow-client#299 updates the workflow client to allow you to send in the new metadata as a hash and it will send it to the workflow-server correctly.  The response from the workflow-server contains the metadata as json in the existing XML response.  This is then parsed the dor-workflow-client to return the metadata as a hash again.

Basically: allow arbitrary metadata (posted as JSON) to be passed in when creating a workflow, and then retrieved when viewing workflows.  Defaults allow the current API to continue to work (backwards compatible).

This makes workflow metadata associated with a druid/version pair, so it follows through all workflows (i.e. can be created when initializing one workflow, and will be available to later workflows too).

Notes:
- no schema is enforced, you can send in any kind of data (as an encoded hash) and it will be stored
- metadata is currently returned with each workflow step ... which is not terribly efficient (since it's the same each time)...
- there is no API call to delete all existing metadata for a druid/version pair (other than sending in an empty hash when creating a new workflow for that druid/version pair)

Todo:
- [x] ~~sanity check with team~~
- [x] ~~deploy to server, see if anything blows up (should be backwards compatible)~~

- Updated/new specs
- Deployed to QA and ran integration tests


justinlittman:
Allowing workflow metadata to be available to all workflows seems to risk unintended consequences. If you set a workflow variable, how will you know what it will affect?

app/controllers/workflows_controller.rb
```diff
@@ -49,7 +49,8 @@ def create
       version: Version.new(
         druid: params[:druid],
         version: params[:version]
-      )
+      ),
+      metadata: params[:metadata]
     ).create_workflow_steps
```
peetucket, line 54: allows you to pass in metadata when creating a new workflow

app/models/workflow_step.rb
```diff
@@ -23,6 +23,10 @@ class WorkflowStep < ApplicationRecord
 
   scope :for_version, ->(version) { where(version:) }
 
+  # On delete: raises an exception if there are is any workflow metadata
+  # Don't destroy because there may be other steps that reference the same druid
+  has_one :workflow_metadata, foreign_key: 'druid', primary_key: 'druid', dependent: :restrict_with_error, inverse_of: :workflow_step
+
```
peetucket, line 29: hmmm...this may be a problem actually.... since when you create a new workflow, we currently destroy any instances of the current one

peetucket:
> Allowing workflow metadata to be available to all workflows seems to risk unintended consequences. If you set a workflow variable, how will you know what it will affect?

The issue with making it workflow specific is that the first workflow created may not be the the one that needs the information contained in the variable.  For example, a user checks a "requires OCR" checkbox in pre-assembly.  Pre-assembly starts `assemblyWF`, but this workflow doesn't care about OCR, even though a later workflow does.  So we need to someone propagate this information down the line of any intermediate workflows until the one that cares about it is initialized.    I'm not sure if there is a better way to this (other than new fields in cocina).

mjgiarlo:
@justinlittman 💬 
> Allowing workflow metadata to be available to all workflows seems to risk unintended consequences. If you set a workflow variable, how will you know what it will affect?

Do you see this as a modeling problem or more of a documentation or specification problem? I.e., would it help if we wrote down which variables would be used when and how, and other values will simply be ignored (or disallowed)?

justinlittman:
Seems like risk could be reduced through documentation and only permitting certain variables.

peetucket:
> Seems like risk could be reduced through documentation and only permitting certain variables.

We could certainly later add some validation to ensure only known variables are being sent.  I think the risk is mitigated ultimately by the fact that most consumers of the workflow service will just not care about these variables.  They will get them in the response but will just ignore them unless they are coded to specifically look for them.

Without validation, I think the biggest risk is that you have a typo somewhere and inadvertently set the variable incorrectly and/or miss that the correct one you are expecting, which will cause weird behavior.

Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'rails', '~> 7.0.0'
 
 gem 'bootsnap', '>= 1.1.0', require: false
 gem 'bunny', '~> 2.17'
+gem 'composite_primary_keys'
```
mjgiarlo, line 10: Oh, shoot, I didn't realize composite primary keys were a Rails 7.1 thing?

app/services/workflow_creator.rb
```diff
@@ -28,6 +28,9 @@ def create_workflow_steps
       processes.map do |process|
         WorkflowStep.create!(workflow_attributes(process))
       end
+
+      # Create/update metadata if passed in
+      version.update_metadata if version.metadata
```
mjgiarlo, line 33: Maybe a six-vs-half-dozen situation, but what would you feel about moving the responsibility for this check into the `VersionMetadata#update_metadata` method: https://github.com/sul-dlss/workflow-server-rails/pull/745/files#diff-61977d248b654178763e77a2fddc09152a20b85b1aecfc8222ef4233564593fcR15 Like, have it no-op if metadata is not present.

db/migrate/20240402194159_create_version_metadatas.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateVersionMetadatas < ActiveRecord::Migration[7.0]
+  def change
+    create_table :version_metadata, primary_key: %i[druid version] do |t|
+      t.string      :druid, null: false, index: true
```
mjgiarlo, line 6: In practice I think we only ever care about druid/version pairs, so you could consider dropping `index: true` here (and from the db schema)?

app/services/workflow_creator.rb
peetucket: yeah thought about that and have no opinion -- not sure if there is a benefit one way or another, unless it's more clear that way

db/migrate/20240402194159_create_version_metadatas.rb
peetucket: ok

spec/factories/workflow_steps.rb
```diff
@@ -15,5 +15,11 @@
       status { 'completed' }
       completed_at { Time.now }
     end
+
+    trait :with_ocr_metadata do
+      after(:create) do |step|
+        create(:version_metadata, druid: step.druid, version: step.version)
+      end
+    end
```
mjgiarlo, line 23: Since we moved away from hanging this info off the workflow step model, I wonder if we might also consider moving this factory (or removing it since you created a version metadata factory)?

app/services/workflow_creator.rb
mjgiarlo: The potential benefit is not needing to add this check again if we wind up adding more spots where metadata is updated, but since we only have the one spot, maybe moot

peetucket: i'll move it

spec/factories/workflow_steps.rb
peetucket: this is allowing me to use the version metadata factory to create some metadata for the step all at once, which is useful for tests where i am verifying that steps with metadata work as expected, e.g. like in this test: https://github.com/sul-dlss/workflow-server-rails/blob/workflow-variables/spec/requests/workflows/all_workflows_for_one_object_spec.rb#L55-L75

mjgiarlo: Sounds good.

db/migrate/20240402194159_create_version_metadatas.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateVersionMetadatas < ActiveRecord::Migration[7.0]
+  def change
+    create_table :version_metadata, primary_key: %i[druid version] do |t|
+      t.string      :druid, null: false
+      t.integer     :version, default: 1
```
jmartin-sul, line 7: should `version` also be `null: false`, or is that implied by it being part of the composite primary key?

i guess if you wanted to avoid the addition of the composite primary key gem (though i'm not personally opposed to it), you could have a typical integer PK, make `druid` and `version` both explicitly non-null, and put a unique index on the columns individually and on the combo of them so you don't have any dupes for the same druid version?

spec/models/version_metadata_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe VersionMetadata do
+  let(:version_metadata) { FactoryBot.create(:version_metadata) }
+  let(:druid) { version_metadata.druid }
+  let(:version) { version_metadata.version }
```
jmartin-sul, line 8: do the `druid` and `version` vars get used at all?  cruft?

jmartin-sul (commented):
didn't get through the whole thing (all the non-spec parts, most of the specs).  but i like the idea, and i like the implementation from what i saw.  left a couple small non-blocking comments.

reminds me a bit of XCOM in airflow.

spec/models/version_metadata_spec.rb
peetucket: Good catch, it is cruft.  Removing.

db/migrate/20240402194159_create_version_metadatas.rb
peetucket: I like this idea of getting rid of the gem.  Let me re-do the migration, add a model level validation to ensure we don't have dupes as well as a db constraint and ensure the tests still pass.

app/models/application_record.rb
```diff
@@ -3,4 +3,15 @@
...
+  # ensure we have a valid druid with prefix
+  def druid_is_valid
+    errors.add(:druid, 'is not valid') unless valid_druid?
+  end
+
+  # @return [boolean]
+  def valid_druid?
+    DruidTools::Druid.valid?(druid, true) && druid.starts_with?('druid:')
+  end
 end
```
peetucket, line 17: we only have two db models and both require druid validation, so move here to avoid code duplication (can further abstract if/when we add more models that do not require this validation)

mjgiarlo: We could also consider a lightweight validator class like we use in a handful of other places (H2, ETD, sul_pub, DSA...), like so:

https://github.com/sul-dlss/hydra_etd/blob/main/app/models/submission.rb#L21

https://github.com/sul-dlss/hydra_etd/blob/main/app/validators/folio_instance_hrid_validator.rb

so like:

```ruby
class DruidValidator < ActiveModel::Validator
  def validate(record)
    return if DruidTools::Druid.valid?(record.druid, true) && record.druid.starts_with?('druid:')

    record.errors.add(:druid, 'is not valid')
  end
end
```

app/models/version.rb
```diff
@@ -1,14 +1,28 @@
...
-  attr_reader :druid, :version_id
+  attr_reader :druid, :version_id, :metadata
+
+  def update_metadata
+    # if no metadata passed in (nil), do nothing
+    return unless metadata
+
+    # if metadata passed in is an empty hash, delete the version metadata record
+    if metadata.empty?
+      VersionMetadata.find_by(druid:, version: version_id)&.destroy
```
mjgiarlo, line 20: I'm fuzzy on whether we should delete these VersionMetadata records or if it's fine to let lines 22-23 update them with empty metadata values. Maybe it doesn't matter?

db/migrate/20240402194159_create_version_metadatas.rb
```diff
@@ -0,0 +1,14 @@
...
+    create_table :version_metadata do |t|
+      t.string      :druid, null: false
+      t.integer     :version, null: false, default: 1
+      t.jsonb       :values, default: {}
+      t.timestamps
+    end
+
+    add_index :version_metadata, %i[druid version], unique: true
+  end
+end
```
mjgiarlo, line 14: Even better than the composite PK solution, nice!

app/models/version.rb
peetucket: Yeah, it's not a big deal either way, you'd end up with an empty hash in that row I think if you don't do this delete... just tending to my general cleanup tendencies

app/models/application_record.rb
peetucket: Sounds good, I'll do this, and add the single validation line explicitly to both models to make it clear

app/models/version.rb
```diff
@@ -1,14 +1,28 @@
...
 
-  attr_reader :druid, :version_id
+  attr_reader :druid, :version_id, :metadata
+
+  def update_metadata
+    # if no metadata passed in (nil), do nothing
+    return unless metadata
+
+    # if json metadata is passed in but is empty, delete the version metadata record
+    if metadata == '{}'
```
mjgiarlo, line 19: Line 9 suggests `metadata` is a hash but line 19 checks if `metadata` is a (particular) string. I'm guessing one or the other could use a tweak.

peetucket: Yes, good catch.  Originally it was a hash, but was later tweaked to be a JSON string to better match what is stored in the database.  This makes for less back and forth conversions from hash to json.  I tweaked the comments to make it clearer.

.rubocop_todo.yml
```diff
@@ -51,6 +56,13 @@ Rails/Date:
     - 'spec/views/workflows/index.xml.builder_spec.rb'
     - 'spec/views/workflows/show.xml.builder_spec.rb'
 
+# Offense count: 1
+# Configuration parameters: Include.
+# Include: app/models/**/*.rb
+Rails/HasManyOrHasOneDependent:
+  Exclude:
+    - 'app/models/workflow_step.rb'
```
jcoyne, line 64: Why is this excluded? I don't see a has_many or has_one added in this PR.

app/controllers/workflows_controller.rb
```diff
@@ -48,7 +48,8 @@ def create
       processes: initial_parser.processes,
       version: Version.new(
         druid: params[:druid],
-        version: params[:version]
+        version: params[:version],
+        metadata: params[:metadata]
```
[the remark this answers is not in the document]
peetucket: I pushed a commit which parses the incoming JSON in the controller as you suggested instead of in the `Version` class (which now accepts a `hash` instead of a JSON string).  I wrote a new spec to show that if you pass in something that is not parseable to the API, it will blow up.    Is that what you were thinking?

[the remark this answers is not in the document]
peetucket: spec added: https://github.com/sul-dlss/workflow-server-rails/pull/745/files#diff-8dbb6de325a7f83cbcaa0ba40bcd48e3fa545771136f9cd8a50d540eee018364R63-R75

justinlittman:
@peetucket Can you please add a description of this feature to the README describing its intended and proper use?

peetucket:
> @peetucket Can you please add a description of this feature to the README describing its intended and proper use?

I just added a section to the README.  Will also add some more to the dor-workflow-client PR README (which is where most consumers will be using this feature from): https://github.com/sul-dlss/dor-workflow-client/pull/299/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

.rubocop_todo.yml
```diff
@@ -1,12 +1,17 @@
...
+# on 2024-04-15 18:28:32 UTC using RuboCop version 1.63.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 2
+# Configuration parameters: AllowedMethods, AllowedPatterns, CountRepeatedAttributes.
+Metrics/AbcSize:
+  Max: 21
```
jcoyne, line 12: Can you add an exception just where you need it rather than making this exemption globally?

README.md
```diff
@@ -112,6 +112,24 @@ GET    /workflow_queue/all_queued
...
 ```
 
+### Workflow Variables
+
+If a workflow or workflows for a particular object require data to be persisted and available between steps, workflow variables can be set.
+These are per object/version pair and thus available to any step in any workflow for a given version of an object once set.
+
+These data are not persisted in Cocina, and are not preserved or available outside of the workflow-service, so they should only be used to persist information used during workflow processing.
+
+To use, pass in a "metadata" parameter as a hash when creating a workflow. The hash can contain any number of key/value pairs of metadata:
```
jcoyne, line 122: Can we pass this metadata into the body of the post request instead? Possibly using a JSON object?

app/controllers/steps_controller.rb
```diff
@@ -67,7 +67,6 @@ def step_not_found_error
   end
 
   # Returns most recent workflow step
-  # rubocop:disable Metrics/AbcSize
```
jcoyne, line 70: I'm not sure why this change was made. Can you revert it?

README.md
jcoyne, line 122: I'd suggest "context" as a name rather than "metadata".  We already have "metadata" and it means other things. We're overloading that term too much.

app/models/version_metadata.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Models optional metadata that is associated with a druid/version pair for any workflow
+class VersionMetadata < ApplicationRecord
```
jcoyne, line 4: Can we rename to VersionContext ? "Metadata" is very overused and I believe it will be confusing to talk about.

.rubocop_todo.yml
peetucket: ok

app/models/version_metadata.rb
peetucket: OK.  Renaming all references to `VersionContext` and attribute used from `metadata` to `context`.  Will update draft PR of `dor-workflow-client` as well.

README.md
peetucket: Yes, switched to POST.   If you send in context, you will need to set `application/json` headers for the workflow create request.  This will be done by `dor-workflow-client` (in my PR).

jcoyne (changes requested):

spec/factories/version_contexts.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :version_context do
+    sequence :druid do |n|
+      "druid:bb123bc#{n.to_s.rjust(4, '0')}" # ensure we always have a valid druid format
```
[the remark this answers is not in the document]
peetucket: Will also adjust in the `workflow_steps` factory which i copied this from

db/migrate/20240402194159_create_version_contexts.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateVersionContexts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :version_contexts do |t|
+      t.string      :druid, null: false
+      t.integer     :version, null: false, default: 1
+      t.jsonb       :values, default: {}
```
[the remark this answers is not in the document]
peetucket: I generally don't like the idea of naming an attribute with something that is already part of the model name, because it then leads to code that looks like `version_context.context`.  Personally I don't like this and find it less readable than something like `version_context.values` or `version_context.data` or whatever.

app/models/version_context.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Models optional context that is associated with a druid/version pair for any workflow
+class VersionContext < ApplicationRecord
+  validates :druid, uniqueness: { scope: :version }
+  validates_with DruidValidator
```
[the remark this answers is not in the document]
peetucket: Didn't know this was a thing, looks good and will adjust ... though this now means we get a hash back out too from the model, which is fine (but which will require me to change how the tests work and possibly how results are parsed by my `dor-workflow-client` PR).  

Though I think we want just the specific attribute: `serialize :values, coder: JSON`

app/controllers/workflows_controller.rb
```diff
@@ -66,6 +67,10 @@ def destroy
 
   private
 
+  def context
+    params[:context] || nil
```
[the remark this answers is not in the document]
peetucket: I guess it could, but reading the body this way gives you a string.  `params[:context]` gives you the hash you want to work with(as `ActionController::Parameters`, nicely encoded in the `context` block of the params, so they don't conflict with any other params on the form).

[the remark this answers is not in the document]
jcoyne: I'm thinking the whole body is the context without having an explicit `context` key.  So alternatively, `params.except(:version, :druid)`.

app/models/workflow_step.rb
```diff
@@ -42,6 +42,18 @@ def maybe_set_completed
     self.completed_at ||= Time.now
   end
 
+  # any associated version context for this step (if it exists) -- note: same for any workflow/step for a given druid/version combination
+  def context
+    VersionContext.find_by(druid:, version:)&.values
```
jcoyne, line 47:
```suggestion
    VersionContext.find_by(druid:, version:)&.context
```

app/models/workflow_step.rb
```diff
@@ -42,6 +42,18 @@ def maybe_set_completed
...
+  end
+
+  # context as json to return to the client (or nil )
+  def context_as_json
+    return nil unless context
+
+    context.to_json
```
jcoyne, line 54: Can't we just return `context` here?

spec/models/version_context_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe VersionContext do
+  let(:version_context) { FactoryBot.create(:version_context) }
+
+  it 'includes the context as a hash' do
+    expect(version_context.values).to eq({ 'requireOCR' => true, 'requireTranscript' => true })
```
jcoyne, line 9:
```suggestion
    expect(version_context.context).to eq({ 'requireOCR' => true, 'requireTranscript' => true })
```

db/schema.rb
```diff
@@ -10,11 +10,19 @@
...
-ActiveRecord::Schema.define(version: 2020_08_11_212454) do
-
+ActiveRecord::Schema[7.0].define(version: 2024_04_02_194159) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "version_contexts", force: :cascade do |t|
+    t.string "druid", null: false
+    t.integer "version", default: 1, null: false
+    t.jsonb "values", default: {}
```
jcoyne, line 20:
```suggestion
    t.jsonb "context", default: {}
```

jcoyne (changes requested):

peetucket:
Tests updated to match how client posts as JSON; some refactoring; remove unnecessary serialization declaration in model

jcoyne (approved):

Issue #738, opened 2024-03-28
Add close item check step to beginning of accession WF


justinlittman:
This step should Honeybadger alert if the item is open. (In the future, this will be changed to raising.)

One option would be to change the existing start-accessioning step so that it is handled by a robot rather than automatically completed.

jcoyne:
I believe this is already done in DSA by ObjectsController#accession

justinlittman:
ObjectsController#accession provides the ability to change version attributes (e.g., description) when closing. I believe this functionality is available to users in Argo, so removing it from the accessioning endpoint has rippling effects.

That being said, I'd like to see versioning in ObjectsController#accession so I will add it as a separate task to be considered later.

lwrubel:
Needs https://github.com/sul-dlss/dor-services-app/issues/4779.

andrewjbtw:
> I believe this functionality is available to users in Argo, so removing it from the accessioning endpoint has rippling effects.

Yes, this is in the close modal on show pages and gives people a chance to update their version description in case they did something different than originally intended.

aaron-collier:
I picked this up but I'm not entirely sure where this check should go (or how it fits in workflow server). I'm happy to pick it back up with some pairing however (understanding we're at the end of most folks day).

mjgiarlo:
Closed by above PRs.

PR #748, opened 2024-04-05
Remove status from start-accession step to spin up a robot

mjgiarlo:
Partnered with https://github.com/sul-dlss/common-accessioning/pull/1171

# Why was this change made?

Connects to #738

# How was this change tested?

CI


mjgiarlo:
Not sure if I need to worry about https://github.com/sul-dlss/workflow-server-rails/wiki/What-happens-when-a-step-is-added-or-removed-from-a-workflow-definition when testing or rolling this out.

mjgiarlo:
Coverage ding looks bogus 😮

PR #757, opened 2024-04-16
add shell definitions for new OCR and caption workflows

### LINKED ISSUE #742: Create new ocr and caption workflow defintions
Two new workflows:
- ocr
- caption

Each starts with three steps:
- start
- do thing
- end

Robots to follow, will be no-op for now.  See sul-dlss/common-accessioning#1167

peetucket:

Fixes #742 ... add new workflow definitions for OCR and captioning

NOTE: it is quite possible that we will need additional steps as we progress through the workcycle, but it is straightforward to add these later, so I'd suggest we get started with these.

Localhost using dor-workflow-client ensuring new workflow definitions are created and progress as expected from one step to the next when completing the previous step

### DIFF
--- a/config/workflows/caption.xml
+++ b/config/workflows/caption.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0"?>
+<workflow-def id="caption">
+  <process name="start-caption">
+    <label>Initiate captioning for the object</label>
+  </process>
+  <process name="caption-create">
+    <prereq>start-caption</prereq>
+    <label>Generate captions for the audio/video content in the object (if requested and possible)</label>
+  </process>
+  <process name="end-caption">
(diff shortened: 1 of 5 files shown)


justinlittman:
Can you use "ocrWF" and "captionWF" to be consistent with existing naming conventions?

peetucket:
> Can you use "ocrWF" and "captionWF" to be consistent with existing naming conventions?

I am agnostic, this was discussed in a thread in Slack: https://stanfordlib.slack.com/archives/C06EVDJV45N/p1713218802781899

Opinion seems split.  I can definitely see the argument for consistency.

peetucket:
Added `WF` suffix for consistency.

edsu (approved):
🚀

PR #771, opened 2024-05-22
add new worflow step for ocr workspace cleanup

peetucket:

Part of sul-dlss/common-accessioning#1212

See https://github.com/sul-dlss/common-accessioning/pull/1249 for robot implementation

HOLD until the robot implementation is released


edsu (approved):

jmartin-sul (approved):
merging since ed also approved, and since i just merged the robot implementatation

Issue #782, opened 2024-06-04
[lower priority but actionable] Add new endpoint for skipping workflow


peetucket:
This PR in common-accessioning adds some logic to the `ocrWF:start-ocr` robot that skips the rest of the workflow based on a given condition, by setting all remaining steps to `skipped` in reverse order.

If we ever need to do this again, it makes sense to add a new API endpoint to workflow-server-rails to do this.  You'd pass in the name of the workflow, and it would set all remaining steps to skipped (optionally skipping a step you pass in, which could be the step making the call).

We would also need to update `dor-workflow-client` to wrap the new API call in a method.  Create a separate ticket or do as part of this work.

See https://github.com/sul-dlss/common-accessioning/pull/1262

When available, then update common-accessioning so that the start-ocr step uses the new endpoint.

jmartin-sul:
* this is lower priority than all the other work cycle work (in progress development and subsequent debugging) since we have a working, if slightly suboptimal, solution right now on common-accessioning `main`
* if someone does pick this up, please don't change the call for this to WFS that's made from common-accessioning while other testing is happening, to keep from confounding results in the home stretch.

jmartin-sul:
i think this is done as of https://github.com/sul-dlss/common-accessioning/issues/1299 ?

PR #787, opened 2024-06-17
add skip all steps endpoint

dnoneill:
see https://github.com/sul-dlss/workflow-server-rails/issues/782
Also needs https://github.com/sul-dlss/dor-workflow-client/pull/317 merge

QA + rspec

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -14,6 +14,21 @@ class StepsController < ApplicationController
     head :no_content
   end
 
+  # Used by ocrWF to skip all steps if files don't exist
+  # Added as endpoint to centralize logic for future use and to allow all updates to be made in one transaction
+  def skip_all
+    query = WorkflowStep.where(druid: params[:druid],
+                               active_version: true,
+                               workflow: params[:workflow])
+    parser = ProcessParser.new(process_from_request_body)
+    WorkflowStep.transaction do
+      query.each do |record|
+        record.update(status: 'skipped', note: parser.to_h[:note])
+      end
+    end
+    head :no_content
+  end
+
   # Update a single WorkflowStep
   # If there are next steps, they are enqueued.
   # rubocop:disable Metrics/AbcSize
--- a/spec/requests/workflows/workflows_skip_all_spec.rb
+++ b/spec/requests/workflows/workflows_skip_all_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Skip all in a workflow' do
+  let(:wf) { FactoryBot.create(:workflow_step) }
+  let(:druid) { wf.druid }
+  let(:workflow_template) { WorkflowTemplateLoader.load_as_xml('accessionWF') }
+
+  before do
+    allow(QueueService).to receive(:enqueue)
+  end
+
+  describe 'POST skip_all' do
+    let(:druid) { 'druid:bb123bb1234' }
+    let(:workflow) { 'accessionWF' }
+
+    let(:headers) { { 'Content-Type' => 'application/xml' } }
+    let(:xml_content) { '<?xml version="1.0"?><process name="skip-all" note="note"/>' }
+
+    context 'when skip workflow is called' do
+      it 'skips all steps in a workflow' do
+        # create workflow steps
+        post "/objects/#{druid}/workflows/#{workflow}?version=1"
+        # skip all steps
+        post("/objects/#{druid}/workflows/#{workflow}/skip-all", params: xml_content, headers:)
+        expect(WorkflowStep.all.map(&:status).uniq).to eq(['skipped'])
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


app/controllers/steps_controller.rb
```diff
@@ -14,6 +14,22 @@ def destroy_all
     head :no_content
   end
 
+  # endpoint only used for ocrWF to skip all steps if files don't exist
+  # Done to make sure we don't have to load the entire cocina in Argo
+  # Added as endpoint to avoid making multiple calls to update_status for all steps in WF
```
peetucket, line 19:
```suggestion
Endpoint currently only used for ocrWF to skip all steps if files don't exist (as of June 2024)
Added as endpoint to centralize logic for future use and to allow all updates to be made in one transaction
```

peetucket (changes requested):
One small suggested change to the comment (including removing the bit about cocina, as that is pretty specific to ocrWF and why we decided on skipping it right after starting it, as it may be confusing to read for someone who is just editing workflow-server-rails later without the context of ocrWF).  I'd suggest putting that comment in soon to be done common-accessioning PR right above the call to this new endpoint

app/controllers/steps_controller.rb
peetucket: In fact, now that I just typef it, I'm not even sure I like my own suggestion :)

It's prone to be wrong if and when we use this endpoint somewhere else in the future and don't update the comment to match.

Perhaps even just something like `Used by ocrWF to skip all steps if files don't exist` or something that is a pointer without being super specific about if anyone else is also using it

spec/requests/workflows/workflows_skip_all_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    let(:headers) { { 'Content-Type' => 'application/xml' } }
+    let(:xml_content) { '<?xml version="1.0"?><process name="skip-all" note="note"/>' }
+
+    context 'when the version is passed without context' do
+      it 'creates new workflows and skips all steps' do
+        expect do
+          post "/objects/#{druid}/workflows/#{workflow}?version=1"
+        end.to change(WorkflowStep, :count).by(10)
+        post("/objects/#{druid}/workflows/#{workflow}/skip-all", params: xml_content, headers:)
+        expect(WorkflowStep.last(10).map(&:status).uniq).to eq(['skipped'])
```
jmartin-sul, line 27: will these be the only 10 steps in the DB at the time this test is running?  if so, you might be able to drop the `last(10)` to make that clearer?  otherwise, it might be good to add an `order` clause, since return order isn't guaranteed in the absence of one, and since relying on the fact that rows are usually returned in order of last update can sometimes lead to intermittently flappy tests).

spec/requests/workflows/workflows_skip_all_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  describe 'POST skip_all' do
+    let(:druid) { 'druid:bb123bb1234' }
+    let(:workflow) { 'accessionWF' }
+
+    let(:headers) { { 'Content-Type' => 'application/xml' } }
+    let(:xml_content) { '<?xml version="1.0"?><process name="skip-all" note="note"/>' }
+
+    context 'when the version is passed without context' do
```
jmartin-sul, line 21: i'm not sure i understand the context description here, especially the "without context" part for.  wondering whether it's copypasta, or if i'm missing something.

jmartin-sul (commented):
a few small possible nitpicks, but nothing for which i'd necessarily hold up merging

app/controllers/steps_controller.rb
```diff
@@ -14,6 +14,19 @@ def destroy_all
...
   end
 
+  def skip_all
+    query = WorkflowStep.where(druid: params[:druid],
+                               active_version: true,
+                               workflow: params[:workflow])
+    parser = ProcessParser.new(process_from_request_body)
+    WorkflowStep.transaction do
+      query.each do |record|
+        record.update(status: 'skipped', note: parser.to_h[:note])
```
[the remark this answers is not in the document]
dnoneill: I tried update_all originally, I also tried update_all to update the time and it caused a rubocop error, so I changed it to this.

spec/requests/workflows/workflows_skip_all_spec.rb
dnoneill: My bad. I copied a previous test and didn't change that context.

app/controllers/steps_controller.rb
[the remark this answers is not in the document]
peetucket: no strong preference from me, i probably would have done an `update_all` too, though i don't know the details of what works and doesn't ... also this way seems fine to me .. i think regardless its important to be in a transaction and set that updated_at column, which this will do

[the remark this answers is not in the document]
jmartin-sul: ah, ok, well in that case this wfm as-is, thanks for the explanation!

app/controllers/steps_controller.rb
jmartin-sul: oh, fair point, i like the idea of making the comment a little less specific so that it's less prone to going stale.

jmartin-sul: and just as i type that, i see an update to the comment 🙂 (updated version lgtm)

jmartin-sul (approved):

peetucket (approved):

PR #812, opened 2024-10-25
Configure separate queues for DSA robots.

justinlittman:
So that specific robots can be run within DSA.

Unit, integration test

### DIFF
--- a/app/models/workflow_step.rb
+++ b/app/models/workflow_step.rb
@@ -6,7 +6,7 @@ class WorkflowStep < ApplicationRecord
   validates :workflow, presence: true
   validates :process, presence: true
   validates :version, numericality: { only_integer: true }
-  validates :status, inclusion: { in: %w[waiting started completed queued error skipped] }
+  validates :status, inclusion: { in: %w[waiting started completed queued error skipped retrying] }
   validates :process, uniqueness: { scope: %w[version workflow druid] }
   validate  :workflow_exists, on: :create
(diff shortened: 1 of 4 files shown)


app/models/workflow_step.rb
```diff
@@ -6,7 +6,7 @@ class WorkflowStep < ApplicationRecord
   validates :workflow, presence: true
   validates :process, presence: true
   validates :version, numericality: { only_integer: true }
-  validates :status, inclusion: { in: %w[waiting started completed queued error skipped] }
+  validates :status, inclusion: { in: %w[waiting started completed queued error skipped retrying] }
```
justinlittman, line 9: The status of a workflow step will be changed to "retrying" will the job is retrying.

app/services/queue_service.rb
```diff
@@ -26,6 +26,8 @@ def enqueue
 
   private
 
+  DSA_ROBOTS = ['Robots::DorRepo::Accession::Publish'].freeze
```
justinlittman, line 29: More will be added here.

app/models/workflow_step.rb
lwrubel: How is "retrying" set as a status? Is that to be implemented?

lwrubel: Oh, I see, that's in lyber-core.

lwrubel (approved):

PR #818, opened 2024-11-04
Adds special handling for Update Marc robot.

justinlittman:
This will allow Sidekiq to be configured to make Update Marc single threaded.

### DIFF
--- a/app/services/queue_service.rb
+++ b/app/services/queue_service.rb
@@ -37,16 +37,22 @@ class QueueService
     'Robots::DorRepo::Release::UpdateMarc'
   ].freeze
 
+  SPECIAL_ROBOTS = {
+    # Special case because this robot can eats up too much memory if more
+    # than one instance is running on a worker box simultaneously
+    'Robots::DorRepo::Assembly::Jp2Create' => 'assemblyWF_jp2',
+    # Special case so that can single thread Folio updates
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

app/services/queue_service.rb
```diff
@@ -37,16 +37,22 @@ def enqueue
     'Robots::DorRepo::Release::UpdateMarc'
   ].freeze
 
+  SPECIAL_ROBOTS = {
+    # Special case because this robot can eats up too much memory if more
+    # than one instance is running on a worker box simultaneously
+    'Robots::DorRepo::Assembly::Jp2Create' => 'assemblyWF_jp2',
+    # Special case so that can single thread Folio updates
+    'Robotos::DorRepo::Release::UpdateMarc' => 'releaseWF_update-marc_dsa'
```
lwrubel, line 45: Typo here @justinlittman

app/services/queue_service.rb
```diff
@@ -37,16 +37,22 @@ def enqueue
     'Robots::DorRepo::Release::UpdateMarc'
```
lwrubel, line 37: Should this line now be deleted?

app/services/queue_service.rb
justinlittman: Domo Origato, Mr Robotos.

PR #827, opened 2024-12-19
add new workflow step to speechToTextWF

peetucket:

Goes with https://github.com/sul-dlss/common-accessioning/pull/1444 to add a new post processing workflow step to speechToTextWF

Please merge the common-accessioning PR first to be sure the robot is ready before merging this

Integration tests on QA

### DIFF
--- a/config/workflows/speechToTextWF.xml
+++ b/config/workflows/speechToTextWF.xml
@@ -15,8 +15,12 @@
     <prereq>stt-create</prereq>
     <label>Move new STT files to workspace</label>
   </process>
-  <process name="update-cocina">
+  <process name="process-files">
     <prereq>stage-files</prereq>
+    <label>Process speech to text files to remove problematic phrases</label>
+  </process>
+  <process name="update-cocina">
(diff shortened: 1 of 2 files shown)


config/workflows/speechToTextWF.xml
```diff
@@ -15,8 +15,12 @@
     <prereq>stt-create</prereq>
     <label>Move new STT files to workspace</label>
   </process>
-  <process name="update-cocina">
+  <process name="process-files">
```
edsu, line 18: This isn't blocking but "process files" sounds pretty generic--maybe a more descriptive name would be helpful? Something like `apply-stoplist`.

edsu (approved):

config/workflows/speechToTextWF.xml
peetucket: meant to hold off for more discussion, but i just merged the other PR which named the actual robot ... so I'm going to go with this for now, we can rename later if the team thinks this name is too generic

PR #832, opened 2025-01-16
Normalize logging

mjgiarlo:
# Why was this change made?

To normalize logging practices across services. To do this, lean on lograge and remove duplicate code added on upgrade to Rails 7.2.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -38,11 +38,6 @@ Rails.application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
-  # Log to STDOUT by default
-  config.logger = ActiveSupport::Logger.new($stdout)
-                                       .tap  { |logger| logger.formatter = Logger::Formatter.new }
-                                       .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
-
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -38,11 +38,6 @@
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   # config.force_ssl = true
 
-  # Log to STDOUT by default
-  config.logger = ActiveSupport::Logger.new($stdout)
-                                       .tap  { |logger| logger.formatter = Logger::Formatter.new }
-                                       .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
-
```
mjgiarlo, line 45: We didn't need to add this in the Rails 7.1 upgrade because this file already has https://github.com/sul-dlss/workflow-server-rails/pull/832/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eL71-L75 which only logs to `STDOUT` if the `RAILS_LOG_TO_STDOUT` env var is turned on.

config/environments/production.rb
```diff
@@ -76,6 +71,7 @@
 
   config.lograge.enabled = true
   config.lograge.base_controller_class = 'ActionController::API'
+  config.lograge.ignore_actions = ['OkComputer::OkComputerController#show']
```
mjgiarlo, line 74: Do we need to see each GET request Nagios sends to the OkComputer controller every minute?

PR #860, opened 2025-05-29
rails 8 upgrade

peetucket:

Patch party May 29 2025

Rails 8 upgrade new rubocops and fixes

CI


app/services/queue_service.rb
```diff
@@ -44,6 +42,8 @@ def enqueue
     'Robots::DorRepo::Release::UpdateMarc' => 'releaseWF_update-marc_dsa'
   }.freeze
 
+  private
+
```
peetucket, line 46: make rubocop happy by moving private declaration below constants

mjgiarlo (approved):

PR #871, opened 2025-07-21
remove deprecated platforms

peetucket:

`[DEPRECATED] Platform :mingw, :x64_mingw is deprecated. Please use platform :windows instead.`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ source 'https://gems.contribsys.com/' do
 end
 
 group :development, :test do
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -24,7 +24,7 @@ source 'https://gems.contribsys.com/' do
 end
 
 group :development, :test do
-  gem 'byebug', platforms: %i[mri mingw x64_mingw]
+  gem 'byebug', platforms: %i[mri windows]
```
mjgiarlo, line 27: I bet you could dump the entire `platforms` bit here since we don't have any windows environments.

